PR #73, opened 2021-08-04
Improve buffer_cast encoding when memref has an identity map

makesource:
In the previous PR (https://github.com/aqjune/mlir-tv/pull/66), we encoded buffer_cast ops using a forall quantifier.
However, using a forall quantifier is expensive in SMT.
But, in many cases, we can find a cheaper encoding for buffer_cast.
If the origin memref is identity-mapped, it can be encoded more efficiently using `z3::lambda`.
Memref with an identity map means its layout is same as that of the origin tensor.

By using lambda, transformations containing complex buffer_cast operations (variable sized tensor, multi-dimension..) are verified much faster than before.
In the previous encoding, even verifying very small memrefs like `memref<3x4xf32>` failed.

### DIFF
--- a/src/memory.cpp
+++ b/src/memory.cpp
@@ -104,7 +104,7 @@ expr SingleArrayMemory::storeArray(
   arrayMaps = z3::store(arrayMaps, bid, stored);
 
   return z3::ule(low, high) && // low <= high (to prevent overflow)
-    z3::ule(high, block.numelem) && // high <= block.numelem
+    z3::ult(high, block.numelem) && // high < block.numelem
     block.writable;
 }
 
@@ -217,7 +217,7 @@ expr MultipleArrayMemory::storeArray(
     });
 
   return z3::ult(low, high) && // low <= high (to prevent overflow)
-    z3::ule(high, getNumElementsOfMemBlock(bid)) && // high <= block.numelem
+    z3::ult(high, getNumElementsOfMemBlock(bid)) && // high < block.numelem
     getWritable(bid);
 }
 
--- a/src/memory.cpp
+++ b/src/memory.cpp
@@ -103,7 +103,7 @@ expr SingleArrayMemory::storeArray(
   auto stored = z3::lambda(idx, z3::ite(cond, arrayVal, currentVal));
   arrayMaps = z3::store(arrayMaps, bid, stored);
 
-  return z3::ule(low, high) && // low <= high (to prevent overflow)
+  return z3::bvadd_no_overflow(offset, size - 1, false) && // to prevent overflow
     z3::ult(high, block.numelem) && // high < block.numelem
     block.writable;
 }
@@ -216,7 +216,7 @@ expr MultipleArrayMemory::storeArray(
       return z3::lambda(idx, z3::ite(cond, arrayVal, currentVal));
     });
 
-  return z3::ult(low, high) && // low <= high (to prevent overflow)
+  return z3::bvadd_no_overflow(offset, size - 1, false) && // to prevent overflow
     z3::ult(high, getNumElementsOfMemBlock(bid)) && // high < block.numelem
     getWritable(bid);
 }
--- a/src/memory.cpp
+++ b/src/memory.cpp
@@ -68,6 +68,24 @@ z3::expr SingleArrayMemory::store(const z3::expr &f32val,
   return z3::ult(idx, block.numelem) && block.writable;
 }
 
+z3::expr SingleArrayMemory::storeArray(
+  const z3::expr &arr, const z3::expr &bid, const z3::expr &offset, const z3::expr &size) {
+  auto lhs = offset;
+  auto rhs = offset + size;
+  auto idx = Index("idx");
+  auto arrayVal = z3::select(arr, idx - lhs);
+
+  auto block = getMemBlock(bid);
+  auto currentVal = z3::select(block.array, idx);
+  auto cond = z3::ule(lhs, idx) && z3::ult(idx, rhs);
+  auto stored = z3::lambda(idx, z3::ite(cond, arrayVal, currentVal));
+  arrayMaps = z3::store(arrayMaps, bid, stored);
+
+  return z3::ule(lhs, rhs) && // lhs < rhs (to prevent overflow)
+    z3::ult(rhs, block.numelem) &&
+    block.writable;
+}
+
 std::pair<z3::expr, z3::expr> SingleArrayMemory::load(
   const z3::expr &bid, const z3::expr &idx) const {
   const auto block = getMemBlock(bid);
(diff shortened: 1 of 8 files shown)


tests/litmus/memref-ops/var_buffer_cast.src.mlir
```diff
@@ -0,0 +1,13 @@
+// EXPECT: "Memory mismatch"
+
+// Although this conversion is correct, currently we don't fully support local block memory refinement.
+// So this test just checks whether return value mismatch.
+
```
makesource, line 5: Maybe local block refinement will be supported soon!

src/value.cpp
```diff
@@ -498,12 +500,30 @@ MemRef::getDimsAndLayoutAndElemTy(
 
 pair<z3::expr, z3::expr> MemRef::load(const vector<z3::expr> &indices) {
   z3::expr idx = to1DIdxWithLayout(indices);
-  return m->load(bid, offset + idx);
+  auto [expr, success] = m->load(bid, offset + idx);
+
+  // check whether indices are  inbound.
+  for (int i = 0; i < indices.size(); i ++)
+    success = success && z3::ult(indices[i], getDim(i));
+
```
makesource, line 508: This may contradicts with https://github.com/aqjune/mlir-tv/pull/32#discussion_r669383881.
But in mlir documents (https://mlir.llvm.org/docs/Dialects/Builtin/#index-map), the index space is always non-negative and integral. So it is not allowed to access with negative Index.

And I think it is fair to mark out-of-bounds access as UB, as out-of-bounds pointer access is UB.
With out this, below `var_buffer_cast.src.mlir`- `var_buffer_cast.tgt.mlir` pair results in "Source is more defined than target" because tensor already checks out-of-bounds access and mark it as UB

src/value.cpp
```diff
@@ -498,12 +500,30 @@ MemRef::getDimsAndLayoutAndElemTy(
...
 
 z3::expr MemRef::store(const z3::expr &value, const std::vector<z3::expr> &indices) {
   z3::expr idx = to1DIdxWithLayout(indices);
-  return m->store(value, bid, offset + idx);
+  auto success = m->store(value, bid, offset + idx);
+
+  // check whether indices are  inbound.
+  for (int i = 0; i < indices.size(); i ++)
+    success = success && z3::ult(indices[i], getDim(i));
+
```
makesource, line 519: same here

src/value.cpp
makesource: Hmm.. it seems like difficult issue.
Maybe this dimension wise out-of-bounds check should be relaxed.

```
int arr[64][64];
return arr[4][100]; // 4*64 + 100 < 64*64, so it is well defined?
```

makesource: Although above expression is admitted in C semantic, dimension-wise index bounds are strictly checked in MLIR. 
`mlir::boundCheckLoadOrStoreOp` in `mlir/lib/Analysis/Utils.cpp` may relate with this.
So we need this bound checks in well-defineness.

makesource:
This PR suspended until local memblocks supported :)

tests/litmus/memref-ops/var_buffer_cast.src.mlir
makesource: Local Memblock is now supported through https://github.com/aqjune/mlir-tv/pull/79.
So mlir-tv correctly verify this litmus tests

src/value.cpp
```diff
@@ -516,12 +518,28 @@ MemRef::getDimsAndLayoutAndElemTy(
 
 pair<expr, expr> MemRef::load(const vector<expr> &indices) {
   expr idx = to1DIdxWithLayout(indices);
-  return m->load(bid, offset + idx);
+  auto [expr, success] = m->load(bid, offset + idx);
+
+  // check whether indices are inbound.
+  for (int i = 0; i < indices.size(); i ++)
+    success = success && z3::ult(indices[i], getDim(i));
+
```
makesource, line 526: This may contradicts with #32 (comment).
But in mlir documents (https://mlir.llvm.org/docs/Dialects/Builtin/#index-map), the index space is always non-negative and integral. So it is not allowed to access with negative Index.

And dimension-wise index bounds are strictly checked in MLIR.
mlir::boundCheckLoadOrStoreOp in mlir/lib/Analysis/Utils.cpp may relate with this.
So we need this bound checks in well-defineness.

src/value.cpp
```diff
@@ -516,12 +518,28 @@ MemRef::getDimsAndLayoutAndElemTy(
...
 
 expr MemRef::store(const expr &value, const std::vector<expr> &indices) {
   expr idx = to1DIdxWithLayout(indices);
-  return m->store(value, bid, offset + idx);
+  auto success = m->store(value, bid, offset + idx);
+
+  // check whether indices are inbound.
+  for (int i = 0; i < indices.size(); i ++)
+    success = success && z3::ult(indices[i], getDim(i));
+
```
makesource, line 537: same here

makesource:
This PR is ready to be reviewed!

src/memory.cpp
```diff
@@ -90,6 +90,24 @@ expr SingleArrayMemory::store(const expr &f32val,
...
+  auto arrayVal = z3::select(arr, idx - low);
+
+  auto block = getMemBlock(bid);
+  auto currentVal = z3::select(block.array, idx);
+  auto cond = z3::ule(low, idx) && z3::ule(idx, high);
+  auto stored = z3::lambda(idx, z3::ite(cond, arrayVal, currentVal));
+  arrayMaps = z3::store(arrayMaps, bid, stored);
+
+  return z3::ule(low, high) && // low <= high (to prevent overflow)
+    z3::ule(high, block.numelem) && // high <= block.numelem
```
aqjune, line 107: Is it ule or ult?

src/value.cpp
```diff
@@ -491,7 +491,9 @@ expr MemRef::getWellDefined() const {
 
 optional<tuple<vector<expr>, MemRef::Layout, z3::sort>>
 MemRef::getDimsAndLayoutAndElemTy(
-    mlir::MemRefType memRefTy, bool freshVarForUnknownSize) {
+    mlir::MemRefType memRefTy,
+    optional<vector<expr>> predefinedDims,
```
aqjune, line 495: Let's make this either `optional<vector<expr>>&&` or `const optional<vector<expr>>&`

src/value.cpp
aqjune: Can we split this change as a separate PR?

src/memory.cpp
makesource: Oh `ult` is right because it is zero based. Thank you for correction!

src/value.cpp
makesource: I think it's enough to pass argument with `std::move` in caller. With move, it doesn't introduce value copy.
https://github.com/aqjune/mlir-tv/pull/73/commits/2d2f9999c3827e7fd300e009052bf2ac1e7f1be7#diff-c12f316dc5036218e6ee70c5052677d9601717720f54845049bbe49a4a8acf89R435

src/memory.cpp
```diff
@@ -90,6 +90,24 @@ expr SingleArrayMemory::store(const expr &f32val,
...
+  auto idx = Index("idx");
+  auto arrayVal = z3::select(arr, idx - low);
+
+  auto block = getMemBlock(bid);
+  auto currentVal = z3::select(block.array, idx);
+  auto cond = z3::ule(low, idx) && z3::ule(idx, high);
+  auto stored = z3::lambda(idx, z3::ite(cond, arrayVal, currentVal));
+  arrayMaps = z3::store(arrayMaps, bid, stored);
+
+  return z3::ule(low, high) && // low <= high (to prevent overflow)
```
makesource, line 106: z3::is_overflow..

src/memory.cpp
```diff
@@ -184,6 +202,25 @@ expr MultipleArrayMemory::store(const expr &f32val,
...
+  auto arrayVal = z3::select(arr, idx - low);
+
+  update(bid, [&](auto ubid) { return &arrays[ubid]; },
+    [&](auto ubid) {
+      auto currentVal = z3::select(arrays[ubid], idx);
+      z3::expr cond = z3::ule(low, idx) && z3::ule(idx, high);
+      return z3::lambda(idx, z3::ite(cond, arrayVal, currentVal));
+    });
+
+  return z3::ult(low, high) && // low <= high (to prevent overflow)
```
makesource, line 219: z3::overflow..

src/vcgen.cpp
```diff
@@ -440,6 +442,14 @@ optional<string> encodeOp(State &st, mlir::memref::BufferCastOp op) {
       get<2>(*dimsAndLayoutAndElemTy),
       true);
 
+  if (memrefTy.getAffineMaps().empty()) {
+    // memref with identity map
+    auto success = memref.storeArray(tensor.asArray(), Index::zero(), tensor.get1DSize());
+    st.wellDefined(success);
+    st.regs.add(op.memref(), move(memref));
+    return {};
+  }
```
makesource, line 451: else

PR #74, opened 2021-08-04
Partially implement smt-switch terms for value types

strikef:
This PR implements smt-switch Terms (`smt::Term` below) for Index, Float, and Integer types.
Currently they are not actually used in verification: `smt::Term` should be implemented for more types in order to do so.

Each type holds a map of Terms, of which the key is `SolverName` enum. This is because we want to provide options for solvers at runtime, and therefore the number of terms a type should hold is not fixed.

Note that this PR includes some of the changes from https://github.com/aqjune/mlir-tv/pull/70. Library linking option must be synced between `mlir-tv` and unittests, and the previous PR included refactoring of CMakeLists to reduce that workload. (Yup I should have split them into separate PRs :( )


strikef:
Disregard due to change in implementation decision

PR #79, opened 2021-08-06
Implement local memory blocks

makesource:
To solve memory mismatch issue in `buffer_cast` operations, we introduce a new concept of memory block. 
That is "local memory block" which can exist only in that block scope.
So there blocks published dynamically at runtime and block size is determined at that time.
For simplicity, local memory blocks are not considered when showing refinement between memory for now.

### DIFF
--- a/src/memory.cpp
+++ b/src/memory.cpp
@@ -12,12 +12,18 @@ static unsigned int ulog2(unsigned int numBlocks) {
   return (unsigned int) ceil(log2(std::max(numBlocks, (unsigned int) 2)));
 }
 
-Memory* Memory::create(unsigned int numBlocks, MemEncoding encoding) {
+static unsigned int norm2(unsigned int number) {
+  return (unsigned int) pow(2, ulog2(number));
+}
+
+Memory* Memory::create(
+    unsigned int globalBlocks, unsigned int localBlocks,
+    MemEncoding encoding) {
   switch(encoding) {
   case MemEncoding::SINGLE_ARRAY:
-    return new SingleArrayMemory(numBlocks);
+    return new SingleArrayMemory(norm2(globalBlocks), norm2(localBlocks));
   case MemEncoding::MULTIPLE_ARRAY:
-    return new MultipleArrayMemory(numBlocks);
+    return new MultipleArrayMemory(norm2(globalBlocks), norm2(localBlocks));
   default:
     llvm_unreachable("Unknown memory encoding");
   }
@@ -38,8 +44,8 @@ SingleArrayMemory::refines(const Memory &other) const {
   return {refinement, {bid, offset}};
 }
 
-SingleArrayMemory::SingleArrayMemory(unsigned int numBlocks):
-  Memory(ulog2(numBlocks), numBlocks),
+SingleArrayMemory::SingleArrayMemory(unsigned int globalBlocks, unsigned int localBlocks):
+  Memory(globalBlocks, localBlocks, ulog2(globalBlocks) + ulog2(localBlocks)),
   arrayMaps(ctx.constant("arrayMaps",
     ctx.array_sort(ctx.bv_sort(bidBits), ctx.array_sort(Index::sort(), Float::sort())))),
   writableMaps(ctx.constant("writableMaps",
@@ -76,9 +82,9 @@ std::pair<expr, expr> SingleArrayMemory::load(
 }
 
 
-MultipleArrayMemory::MultipleArrayMemory(unsigned int numBlocks):
-    Memory(ulog2(numBlocks), numBlocks) {
-  for (unsigned i = 0; i < numBlocks; ++i) {
+MultipleArrayMemory::MultipleArrayMemory(unsigned int globalBlocks, unsigned int localBlocks):
+    Memory(globalBlocks, localBlocks, ulog2(globalBlocks) + ulog2(localBlocks)) {
+  for (unsigned i = 0; i < getNumBlocks(); ++i) {
     auto suffix = [&](const string &s) {
       return s + to_string(i);
     };
@@ -91,7 +97,7 @@ MultipleArrayMemory::MultipleArrayMemory(unsigned int numBlocks):
 
 expr MultipleArrayMemory::itebid(
     const expr &bid, function<expr(unsigned)> fn) const {
-  assert(numBlocks > 0);
+  assert(getNumBlocks() > 0);
   assert(bid.get_sort().is_bv() && bid.get_sort().bv_size() == getBIDBits());
 
   uint64_t const_bid;
@@ -99,7 +105,7 @@ expr MultipleArrayMemory::itebid(
     return fn(const_bid);
 
   const unsigned bits = bid.get_sort().bv_size();
-  unsigned curbid = numBlocks - 1;
+  unsigned curbid = getNumBlocks() - 1;
   expr val = fn(curbid);
 
   while (curbid) {
@@ -113,7 +119,7 @@ expr MultipleArrayMemory::itebid(
 void MultipleArrayMemory::update(
     const expr &bid, function<expr*(unsigned)> getExprToUpdate,
     function<expr(unsigned)> getUpdatedValue) const {
-  assert(numBlocks > 0);
+  assert(getNumBlocks() > 0);
   assert(bid.get_sort().is_bv() && bid.get_sort().bv_size() == getBIDBits());
 
   uint64_t const_bid;
@@ -123,7 +129,7 @@ void MultipleArrayMemory::update(
   }
 
   const unsigned bits = getBIDBits();
-  for (unsigned i = 0; i < numBlocks; ++i) {
+  for (unsigned i = 0; i < getNumBlocks(); ++i) {
     expr *expr = getExprToUpdate(i);
     assert(expr);
     *expr = z3::ite(bid == ctx.bv_val(i, bits), getUpdatedValue(i), *expr);
@@ -154,7 +160,7 @@ expr MultipleArrayMemory::store(const expr &f32val,
 
 std::pair<expr, expr> MultipleArrayMemory::load(
     unsigned ubid, const expr &idx) const {
-  assert(ubid < numBlocks);
+  assert(ubid < getNumBlocks());
 
   expr success = z3::ult(idx, getNumElementsOfMemBlock(ubid));
   return {z3::select(arrays[ubid], idx), success};
@@ -175,7 +181,7 @@ MultipleArrayMemory::refines(const Memory &other0) const {
   // a plain LLVM.
   const MultipleArrayMemory &other =
       *static_cast<const MultipleArrayMemory *>(&other0);
-  assert(other.numBlocks == numBlocks);
+  assert(other.getNumBlocks() == getNumBlocks());
 
   auto bid = ctx.bv_const("bid", bidBits);
   auto offset = Index("offset", true);
--- a/src/memory.cpp
+++ b/src/memory.cpp
@@ -51,8 +51,7 @@ SingleArrayMemory::SingleArrayMemory(
     ctx.array_sort(ctx.bv_sort(bidBits), Index::sort()))),
   isGlobalMaps(ctx) {
   auto bid = ctx.bv_const("bid", bidBits);
-  isGlobalMaps = z3::lambda(bid,
-    z3::ite(z3::ult(bid, numGlobalBlocks), ctx.bool_val(true), ctx.bool_val(false)));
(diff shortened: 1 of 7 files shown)


src/value.cpp
```diff
@@ -449,16 +449,38 @@ expr Tensor::to1DArrayWithOfs(
...
+    bid = m->addLocalMemBlock(get1DSize());
+    offset = Index::zero();
+  } else {
+    static int count = 0;
+    string freshName =  + "memref" + to_string(count ++);
+    bid = ctx.bv_const((freshName + "_bid").c_str(), m->getBIDBits());
+    offset = Index((freshName + "_offset").c_str());
+  }
+}
+
```
makesource, line 467: always create new fresh name for "bid", "offset" when memref name is not specified.

src/vcgen.cpp
```diff
@@ -432,17 +433,18 @@ optional<string> encodeOp(State &st, mlir::memref::BufferCastOp op) {
   if (!dimsAndLayoutAndElemTy)
     return "unsupported type";
 
-  auto memref = MemRef(st.m.get(), "memref",
+  auto memref = MemRef(st.m.get(),
       get<0>(*dimsAndLayoutAndElemTy),
       get<1>(*dimsAndLayoutAndElemTy),
-      get<2>(*dimsAndLayoutAndElemTy));
+      get<2>(*dimsAndLayoutAndElemTy),
+      true);
```
makesource, line 441: fix bugs which "BufferCast" op always introduces same "bid", "offset" z3 variable

makesource, line 441: fix bugs which "BufferCast" op always introduces same "bid", "offset" z3 variable

src/memory.cpp
```diff
@@ -12,17 +12,31 @@ static unsigned int ulog2(unsigned int numBlocks) {
   return (unsigned int) ceil(log2(std::max(numBlocks, (unsigned int) 2)));
 }
 
-Memory* Memory::create(unsigned int numBlocks, MemEncoding encoding) {
+static unsigned int norm2(unsigned int number) {
```
aqjune, line 15: Hm... what is the semantics of this function?
Let's clarify it as a comment.

src/memory.cpp
```diff
@@ -12,17 +12,31 @@ static unsigned int ulog2(unsigned int numBlocks) {
...
+  return (unsigned int) pow(2, ulog2(number));
+}
+
+Memory* Memory::create(
+    unsigned int numGlobalBlocks, unsigned int numLocalBlocks,
+    MemEncoding encoding) {
   switch(encoding) {
   case MemEncoding::SINGLE_ARRAY:
-    return new SingleArrayMemory(numBlocks);
+    return new SingleArrayMemory(norm2(numGlobalBlocks), norm2(numLocalBlocks));
```
aqjune, line 24: Why do we want to normalize it?
It will increase the number of blocks, no?

src/memory.cpp
```diff
@@ -12,17 +12,31 @@ static unsigned int ulog2(unsigned int numBlocks) {
...
   case MemEncoding::MULTIPLE_ARRAY:
-    return new MultipleArrayMemory(numBlocks);
+    return new MultipleArrayMemory(norm2(numGlobalBlocks), norm2(numLocalBlocks));
   default:
     llvm_unreachable("Unknown memory encoding");
   }
 }
 
+expr Memory::isGlobalBlock(const expr &bid) const {
```
aqjune, line 32: We can check whether a bid is non-local by masking MSB and comparing it to zero.
ult is expensive, maybe?

src/memory.cpp
makesource: It finds the minimum power of two value greater than equal to input.
I’ll update it in comments

src/memory.cpp
makesource: This is because I assumed the number of global blocks and local blocks could be different. Is it too much assumption..?

aqjune: I agree they can be different, but is the local/nonlocal bid distinguished by MSB?
Or maybe we can try a different encoding which is suggested in this patch..? what do you think?

makesource: Aha then separating `isLocal` bit from bid bits is a good idea! I'll update this :)

src/memory.cpp
makesource: Well if we separate `isLocal` bit from bids, then there's no need to normalize it. I'll remove this

makesource:
I wonder if I understood correctly and changed it..
I got some interesting results for running time comparing before and after the changes.

## Before changes (type: SINGLE_ARRAY)

```
92% tests passed, 2 tests failed out of 26

Total Test time (real) =  11.05 sec
```

## Before changes (type: MULTIPLE_ARRAY)

```
92% tests passed, 2 tests failed out of 26

Total Test time (real) =  10.54 sec
```

## After changes (type: SINGLE_ARRAY)

```

92% tests passed, 2 tests failed out of 26

Total Test time (real) =  13.09 sec
```

## After changes (type: MULTIPLE_ARRAY)

```
92% tests passed, 2 tests failed out of 26

Total Test time (real) =  11.07 sec
```

aqjune:
That amount of difference is fine, no worries.

src/memory.cpp
```diff
@@ -35,17 +37,23 @@ SingleArrayMemory::refines(const Memory &other) const {
...
     ctx.array_sort(ctx.bv_sort(bidBits), ctx.array_sort(Index::sort(), Float::sort())))),
   writableMaps(ctx.constant("writableMaps",
     ctx.array_sort(ctx.bv_sort(bidBits), ctx.bool_sort()))),
   numelemMaps(ctx.constant("numelemMaps",
-    ctx.array_sort(ctx.bv_sort(bidBits), Index::sort()))) {}
+    ctx.array_sort(ctx.bv_sort(bidBits), Index::sort()))),
+  isGlobalMaps(ctx) {
+  auto bid = ctx.bv_const("bid", bidBits);
+  isGlobalMaps = z3::lambda(bid,
+    z3::ite(z3::ult(bid, numGlobalBlocks), ctx.bool_val(true), ctx.bool_val(false)));
```
aqjune, line 55: What about this?
```
isGlobalMaps = [](const smt::expr &bid) {
  return z3::ult(bid, numGlobalBlocks);
};
```

aqjune: This simplifies the encoding by our own.

src/memory.cpp
```diff
@@ -76,44 +100,50 @@ std::pair<expr, expr> SingleArrayMemory::load(
...
+  Memory(numGlobalBlocks, nunLocalBlocks, ulog2(numGlobalBlocks + nunLocalBlocks)) {
+  for (unsigned i = 0; i < getNumBlocks(); ++i) {
     auto suffix = [&](const string &s) {
       return s + to_string(i);
     };
     arrays.push_back(ctx.constant(suffix("array").c_str(),
         ctx.array_sort(Index::sort(), Float::sort())));
     writables.push_back(ctx.bool_const(suffix("writable").c_str()));
     numelems.push_back(ctx.bv_const(suffix("numelems").c_str(), Index::BITS));
+    isGlobals.push_back(ctx.bool_val(i < numGlobalBlocks));
```
aqjune, line 114: BTW, let's leave this scheme as a comment at somewhere:
```
// Bids smaller than numGlobalBlocks are global
```

src/memory.cpp
```diff
@@ -76,44 +100,50 @@ std::pair<expr, expr> SingleArrayMemory::load(
...
+  for (unsigned i = 1; i < numGlobalBlocks; i ++)
+    globalExpr = z3::ite(bid == ctx.bv_val(i, bits), fn(i), globalExpr);
+
+  if (currLocalBlocks == 0)
+    return globalExpr;
 
-  return val;
+  expr localExpr = fn(numGlobalBlocks);
+  for (unsigned i = numGlobalBlocks + 1; i < getNumBlocks(); i ++)
+    localExpr = z3::ite(bid == ctx.bv_val(i, bits), fn(i), localExpr);
```
aqjune, line 138: Hmm.. do we need to separate these two loops? Maybe we keep them as one as before.

src/memory.h
```diff
@@ -16,28 +16,41 @@ class MemBlock {
...
-  MemBlock(const smt::expr &array, const smt::expr &writable,
-           const smt::expr &numelem):
+  MemBlock(const smt::expr &array, const smt::expr &writable, const smt::expr &numelem):
     array(array), writable(writable), numelem(numelem) {}
 };
 
 class Memory {
 protected:
+  const unsigned int numGlobalBlocks;
+  const unsigned int numLocalBlocks;
```
aqjune, line 26: Let's rename this to `maxLocalBlocks`, and rename `currLocalBlocks` to `numLocalBlocks`.

src/memory.cpp
makesource: At first, I think it's helpful when encoding bid which pointing local memblock (because #local memblock is usually smaller than #global memblock). 
But considering `isGlobalBlock(bid)` is also expressed by long z3::ite terms, my thought might not be true.
I'll fix it using single loop.

src/memory.cpp
makesource: I added it https://github.com/aqjune/mlir-tv/pull/79/files#diff-3d8d4950551437b1a628fd00613869df6939a434606baa3cf36bd391aa463154R49-R52.

src/memory.cpp
makesource: I update it as `isGlobalMaps = z3::lambda(bid, z3::ult(bid, numGlobalBlocks));`

src/memory.cpp
```diff
@@ -123,13 +145,33 @@ void MultipleArrayMemory::update(
...
 
+expr MultipleArrayMemory::isGlobalBlock(const expr &bid) const {
+  return itebid(bid, [&](auto ubid) { return isGlobals[ubid]; });
+}
+
+expr MultipleArrayMemory::isLocalBlock(const expr &bid) const {
+  return itebid(bid, [&](auto ubid) { return !isGlobals[ubid]; });
+}
+
+expr MultipleArrayMemory::addLocalMemBlock(const expr &numelem) {
```
makesource, line 163: addLocalBlock

makesource: add writable argument

src/memory.h
```diff
@@ -16,28 +16,43 @@ class MemBlock {
   smt::expr writable; // bool::sort()
   smt::expr numelem;  // Index::sort()
 
-  MemBlock(const smt::expr &array, const smt::expr &writable,
-           const smt::expr &numelem):
+  MemBlock(const smt::expr &array, const smt::expr &writable, const smt::expr &numelem):
```
makesource, line 19: diff

src/value.cpp
```diff
@@ -449,27 +449,50 @@ expr Tensor::to1DArrayWithOfs(
 
 MemRef::MemRef(Memory *m): m(m), bid(ctx), offset(ctx), layout(Layout({}, ctx)) {}
 
+MemRef::MemRef(Memory *m,
+    const std::vector<expr> &dims,
+    const Layout &layout,
+    const z3::sort &elemty,
+    bool freshBlock): m(m), bid(ctx), offset(ctx), dims(dims), layout(layout) {
+  if (freshBlock) {
```
makesource, line 457: call below initializer

src/value.cpp
```diff
@@ -449,27 +449,50 @@ expr Tensor::to1DArrayWithOfs(
...
-    layout(layout) {}
+    layout(layout) {
+  if (freshBlock) {
+    bid = m->addLocalMemBlock(get1DSize());
+    offset = Index::zero();
+  }
+}
 
 expr MemRef::getWellDefined() const {
+  expr wellDefined = z3::ult(bid, m->getNumBlocks()); // memref points currently issued memblock.
```
makesource, line 486: remove?

src/memory.cpp
```diff
@@ -54,6 +66,15 @@ MemBlock SingleArrayMemory::getMemBlock(const expr &bid) const {
   return MemBlock(array, writable, numelem);
 }
 
+expr SingleArrayMemory::addLocalBlock(const expr &numelem, const expr &writable) {
+  assert(numLocalBlocks < maxLocalBlocks);
```
makesource, line 70: New local block can be created only when numLocalBlocks is sticky lower than maxLocalBlocks.

makesource:
After optimization regarding to `isGlobalBlock`,
`SINGLE_ARRAY` memType checks all tests with `Total Test time (real) =  10.65 sec`,
`MULTIPLE_ARRAY` memType checks all tests with `Total Test time (real) =  10.93 sec`

PR #80, opened 2021-08-07
Fix compile failures in latest c++ stdlibs

strikef:
This PR fixes compile failures when using the latest stdlib (to be precise, libstdc++ 11.2)

Some files that use `std::optional<T>` were missing header `<optional>`. This still worked in the older stdlibs, but it seems the libraries have since been refactored and doesn't compile anymore.
Resolved this issue by simply adding headers in each file.

Also, the `MLIRInferTypeOpInterface` and `MLIRViewLikeInterface` library were written twice in CMakeLists. Deduped both libraries.

### DIFF
--- a/src/value.h
+++ b/src/value.h
@@ -6,7 +6,6 @@
 #include <string>
 #include <optional>
 #include <vector>
-#include <concepts>
 
 class Memory;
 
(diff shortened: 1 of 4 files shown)


src/value.h
```diff
@@ -4,7 +4,9 @@
 #include "llvm/ADT/APFloat.h"
 #include "mlir/Dialect/Linalg/IR/LinalgOps.h"
 #include <string>
+#include <optional>
 #include <vector>
+#include <concepts>
```
aqjune, line 9: Is this needed?

strikef: Oops... forgot to remove this while testing some changes

PR #82, opened 2021-08-09
Update CMakeLists

strikef:
This PR makes a lot of changes in CMakeLists to prevent possible misconfigurations and support more diverse environments.
It is also a prelude for supporting multiple smt solvers.

- Refactored global functions into target-specific functions
- Added options for using libc++
- cmake now fails if one or more necessary arguments aren't provided
- Removed CACHE STRING
- Removed redundant -fPIC option
- Updated README accordingly


strikef:
Temporarily closing the PR due to bugfix

PR #83, opened 2021-08-09
Add encoding memref.subview operation

makesource:
**(This PR contains changes of https://github.com/aqjune/mlir-tv/pull/73)**

In this PR, we support `memref.subview` encoding.
Memref subview operation makes an alias object of original memref with different offset, sizes, strides.

The detailed subview behavior can be found in [documents](https://mlir.llvm.org/docs/Dialects/MemRef/#memrefsubview-mlirmemrefsubviewop).

In documents, there's comments about how the bounds of result memref is defined.
```
// Note that the subview op does not guarantee that the result
// memref is "inbounds" w.r.t to base memref. It is upto the client
// to ensure that the subview is accessed in a manner that is
// in-bounds.
```
To carefully deal with this inbounds check, the every memref's `Layout` should carry this inbounds information.
Then the bounds of resulting memref are determined from the source's inbounds information.

After this PR, all `fold-memref-subview-op` optimizations are verified by mlir-tv! 

스크린샷 2021-08-09 오후 6 37 09


src/vcgen.cpp
```diff
@@ -451,6 +462,84 @@ optional<string> encodeOp(State &st, mlir::memref::BufferCastOp op) {
...
+    z3::expr reducedInbounds = layout.inbounds
+      .substitute(toExprVector(substIndVars), toExprVector(substConsts));
+    auto reducedLayout = MemRef::Layout(reducedIndVars, reducedExpr, reducedInbounds);
+    auto memref = MemRef(st.m.get(), src.getBID(), src.getOffset(),
+      reducedSizes, reducedLayout, Float::sort());
+    st.regs.add(op.getResult(), move(memref));
+
+  } else {
+    auto memref = MemRef(st.m.get(), src.getBID(), src.getOffset(),  sizes, layout, Float::sort());
+    st.regs.add(op.getResult(), move(memref));
```
makesource, line 538: Here we pass source memref's bid, offset expression.
So the result memref has same variable of source's which makes aliasing result.

aqjune:
Hmm.. this PR is pretty big.

Do you think this PR can be splitted into smaller ones?

makesource:
Okay. I'll split it into 2~3 sub pull requests.
Each PR may relate with..
- Subview layout calculation
- Inbounds check for subviewed memref
- Subview rank reduction

PR #86, opened 2021-08-10
Encode subview parameters and calculate transformed layout

makesource:
In this PR, we support memref.subview encoding.
Memref subview operation makes an alias object of original memref with different offset, sizes, strides.

The detailed subview behavior can be found in [documents](https://mlir.llvm.org/docs/Dialects/MemRef/#memrefsubview-mlirmemrefsubviewop).

The basic calculation logic is like this. (offsets, sizes, strides are parameters of subview operation)
```
// Before : <(d0, d1) -> (d0 * s0 + d1)>,
// After: <(d0, d1) -> ((d0 + offsets[0]) * strides[0] * s0 + (d1 + offsets[1]) * strides[1])>
```

The result memref layout information are also doubly checked by MLIR verifier whether it matches with expected value.
(see `Type SubViewOp::inferResultType` in MLIR code)

### DIFF
--- a/src/value.cpp
+++ b/src/value.cpp
@@ -564,9 +564,9 @@ void MemRef::setWritable(bool writable) {
   m->setWritable(bid, writable);
 }
 
-MemRef MemRef::subview(const std::vector<z3::expr> &offsets,
-    const std::vector<z3::expr> &sizes,
-    const std::vector<z3::expr> &strides) {
+MemRef MemRef::subview(const vector<expr> &offsets,
+    const vector<expr> &sizes,
+    const vector<expr> &strides) {
   auto layout = createSubViewLayout(offsets, strides);
   auto memref = MemRef(m, bid, offset, sizes, layout, Float::sort());
   return memref;
@@ -604,14 +604,14 @@ expr MemRef::to1DIdxWithLayout(const vector<expr> &idxs) {
 }
 
 MemRef::Layout MemRef::createSubViewLayout(
-   const vector<z3::expr> &offsets,
-   const vector<z3::expr> &strides) {
+   const vector<expr> &offsets,
+   const vector<expr> &strides) {
    // Before : <(d0, d1) -> (d0 * s0 + d1)>,
    // After: <(d0, d1) -> ((d0 + offsets[0]) * strides[0] * s0 + (d1 + offsets[1]) * strides[1])>
    assert(layout.indVars.size() == offsets.size());
    assert(layout.indVars.size() == strides.size());
 
-   vector<z3::expr> idxs;
+   vector<expr> idxs;
    for (unsigned i = 0; i < layout.indVars.size(); i ++)
      idxs.push_back((layout.indVars[i] + offsets[i]) * strides[i]);
 
--- a/src/value.cpp
+++ b/src/value.cpp
@@ -564,6 +564,23 @@ void MemRef::setWritable(bool writable) {
   m->setWritable(bid, writable);
 }
 
+MemRef::Layout MemRef::toSubViewLayout(
+   const vector<z3::expr> &offsets,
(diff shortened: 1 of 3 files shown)


src/vcgen.cpp
```diff
@@ -426,6 +426,55 @@ optional<string> encodeOp(State &st, mlir::memref::StoreOp op) {
...
+optional<string> encodeOp(State &st, mlir::memref::SubViewOp op) {
+  auto sizes = getSizes(st, op);
+  auto offsets = getOffsets(st, op);
+  auto strides = getStrides(st, op);
+  auto src = st.regs.get<MemRef>(op.source());
+  auto layout = src.toSubViewLayout(offsets, strides);
+
+  auto memref = MemRef(st.m.get(), src.getBID(), src.getOffset(),  sizes, layout, Float::sort());
+  st.regs.add(op.getResult(), move(memref));
+  return "Unsupported yet..";
```
makesource, line 475: Although we support basic layout encoding in SubviewOp, proper inbounds check is not supported yet.
So we leave it unsupported for now..

src/vcgen.cpp
```diff
@@ -426,6 +426,55 @@ optional<string> encodeOp(State &st, mlir::memref::StoreOp op) {
   return {};
 }
 
+vector<z3::expr> getSizes(const State &st, mlir::memref::SubViewOp op) {
```
aqjune, line 429: Let's use smt::expr from now

aqjune: I'll directly edit this, no worries

src/value.cpp
```diff
@@ -564,6 +564,23 @@ void MemRef::setWritable(bool writable) {
   m->setWritable(bid, writable);
 }
 
+MemRef::Layout MemRef::toSubViewLayout(
+   const vector<z3::expr> &offsets,
+   const vector<z3::expr> &strides) {
+   // Before : <(d0, d1) -> (d0 * s0 + d1)>,
+   // After: <(d0, d1) -> ((d0 + offsets[0]) * strides[0] * s0 + (d1 + offsets[1]) * strides[1])>
```
aqjune, line 571: I have two questions:
1. Do you have any idea why it is not `((d0 + offsets[0] * strides[0]) * s0 + (d1 + offsets[1] * strides[1]))`?
2. The document says: `strides: memref-rank number of strides that compose multiplicatively with the base memref strides in each dimension.` Is this consistent with your implementation?

makesource: As mentioned in second question, strides should "compose multiplicatively with the base memref strides".
In above example, d0's original stride is "s0" and d1's original stride is "1".
So after subview operation, these strides should be multiplied with strides[0], strides[1] values, resulting in "s0 * strides[0]" as d0's and "1 * strides[1]" as d1's strides

src/value.cpp
```diff
@@ -564,6 +564,23 @@ void MemRef::setWritable(bool writable) {
   m->setWritable(bid, writable);
 }
 
+MemRef::Layout MemRef::toSubViewLayout(
```
makesource, line 567: create

src/vcgen.cpp
```diff
@@ -426,6 +426,30 @@ optional<string> encodeOp(State &st, mlir::memref::StoreOp op) {
...
+    ADD(sizes, Size);
+    ADD(offsets, Offset);
+    ADD(strides, Stride);
+#undef ADD
+  }
+
+  auto src = st.regs.get<MemRef>(op.source());
+  auto layout = src.toSubViewLayout(offsets, strides);
+
+  auto memref = MemRef(st.m.get(), src.getBID(), src.getOffset(),  sizes, layout, Float::sort());
```
makesource, line 448: add MemRef subview method?

PR #88, opened 2021-08-10
Encode memref inbounds check for SubViewOp when load, store memory

makesource:
In this PR, we added inbounds check for SubViewOp
In documents, there's comments about how memref bounds are defined.

```
// Note that the subview op does not guarantee that the result
// memref is "inbounds" w.r.t to base memref. It is upto the client
// to ensure that the subview is accessed in a manner that is
// in-bounds.
```

To carefully deal with this inbounds check, the every memref's Layout should carry this inbounds information.
Then result memref bounds are determined from the source's memref information. 

For example,

```C
%0: memref<2x3xf32>
// %0's Layout is defined..
// layout.indVars = d0, d1
// layout.expr = d0 * 3 + d1
// layout.inbounds = d0 < 2 && d1 < 3

%1 = memref.subview %0[0, 0][2, 2][2, 1] : 
    memref<2x3xf32> to memref<2x2xf32, affine_map<(d0, d1) -> (d0 * 3 * 2 + d1)>>
// %1's Layout is defined..
// layout.indVars = d0, d1
// layout.expr = d0 * 6 + d1
// layout.inbounds = d0 * 2 < 2 && d1 < 3

%2 = memref.load %1[1, 1] : memref<2x2xf32, affine_map<(d0, d1) -> (d0 * 3 * 2 + d1)>>
// The inbounds condition is calculated like this,
// %1.layout.inbounds.substitude(indVars, {1, 1}) = 1 * 2 < 2 && 1 < 3 = false
// So this load operation is not well defined. 
```

Here we assume `strides`, `offsets` operands are positive values.
There's no explicit descriptions in documents, we can find the motivation of strided memref in [Strided MemRef Proposal](https://groups.google.com/a/tensorflow.org/g/mlir/c/MaL8m2nXuio/).
```
...
A stride specification is a list of integer values that are either non-negative (static case) or `-1` (dynamic case).
Strides encode the distance in memory between successive entries along a particular dimension.
...
```

### DIFF
--- a/src/value.cpp
+++ b/src/value.cpp
@@ -46,16 +46,19 @@ getLayout(const mlir::MemRefType &memRefTy, const vector<expr> &dims) {
     expr layout = Index::zero();
     expr stride = Index::one();
     vector<expr> indVars;
+    expr inbounds = ctx.bool_val(true);
 
-    for (int i = 0; i < dims.size(); i ++)
+    for (int i = 0; i < dims.size(); i ++) {
       indVars.push_back(Index("idx" + to_string(i)));
+      inbounds = inbounds && z3::ult(indVars[i], dims[i]);
+    }
 
     for (int i = dims.size() - 1; i >= 0; i --) {
       layout = layout + stride * indVars[i];
       stride = stride * dims[i];
     }
 
-    return MemRef::Layout(indVars, layout);
+    return MemRef::Layout(indVars, layout, inbounds);
   } else {
     int64_t offset;
     llvm::SmallVector<int64_t, 4> strides;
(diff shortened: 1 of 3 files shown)


src/value.cpp
```diff
@@ -517,25 +522,17 @@ MemRef::getDimsAndLayoutAndElemTy(
...
+  auto [idx, inbounds] = to1DIdxWithLayout(indices);
   auto [expr, success] = m->load(bid, offset + idx);
 
-  // check whether indices are inbound.
-  for (int i = 0; i < indices.size(); i ++)
-    success = success && z3::ult(indices[i], getDim(i));
-
-  return {expr, success.simplify()};
+  return {expr, (success && inbounds).simplify()};
 }
```
makesource, line 530: The original inbounds check is replaced with `inbounds` information from to1DIdxWithLayout.

PR #91, opened 2021-08-10
Implement prototype smt::Expr

strikef:
This PR presents a proof-of-concept expr wrapper class, `smt::Expr`.

This class has three major goals as of now:
- Hide away solver-specific interface as much as possible
- Prevent potential performance degradation while dealing with many expressions
- Reduce the complexity of supporting new solvers

This class doesn't have enough methods to be fully functional, so it hasn't replaced any `z3::expr`s yet.

### DIFF
--- a/src/smt.h
+++ b/src/smt.h
@@ -28,40 +28,15 @@ std::string or_omit(const expr &e);
 
 class Expr {
 private:
-    std::optional<z3::expr> z3_expr;
+  std::optional<z3::expr> z3_expr;
 
-    bool checkZ3(const Expr& arg0) {
-        return arg0.z3_expr.has_value();
-    }
-
-    template<typename... Es>
-    bool checkZ3(const Expr& arg0, const Es&... args) {
-        return arg0.z3_expr.has_value() && checkZ3(args...);
-    }
-
-    template<typename F, typename... Ts>
-    void applyZ3Op(F&& op, const Expr& arg0, const Ts... args) {
-        if (checkZ3(arg0, args...)) {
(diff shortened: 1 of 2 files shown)


src/smt.h
```diff
@@ -15,14 +17,43 @@ std::vector<expr> from1DIdx(
...
+
+public:
+    // default constructor
+    Expr();
+    // move constructor
+    Expr(Expr&& from);
+    // explicit copy 
+    Expr clone() const;
+    // move assignment operator
+    Expr& operator=(Expr&& from);
```
strikef, line 43: Change to = default?

src/smt.h
```diff
@@ -15,14 +17,43 @@ std::vector<expr> from1DIdx(
...
+    // simplify expressions
+    Expr simplify() const;
+    // equivalent to from1DIdx
+    std::vector<Expr> toElements(const std::vector<Expr>& dims) const;
+
+    // update internal z3::expr and get previous z3::expr
+    std::optional<z3::expr> replaceExpr(z3::expr&& z3_expr);
```
strikef, line 50: Implement Expr(solver1::expr, solver2::expr, ...)
Move this method to private?

src/smt.h
```diff
@@ -15,14 +17,43 @@ std::vector<expr> from1DIdx(
...
+
+    friend Expr urem(const Expr& lhs, const Expr& rhs);
+    friend Expr udiv(const Expr& lhs, const Expr& rhs);
```
strikef, line 53: Change to member function

PR #92, opened 2021-08-11
Support Sparse Tensor Constant

KKimj:
## Support Sparse Tensor Constant

It works with ```tensor.extract``` as well.

#### encode ```ConstantOp```
1. check the ```ConstantOp op``` isa ```SparseElementsAttr```
2. check whether  Type of elements, ```eltType``` is ```FloatType``` or ```IntegerType```
    - else ;  ```unsupported type```
3. make Tensor  as invoking ```st.regs.add```  with ```vector``` of ```z3::expr``` and ```dimvec```


KKimj:
Should I write docstring at that code?

makesource:
Could you add simple litmus tests that can check correctness of sparse encoding?

KKimj:
@makesource 
I uploaded tests and resolved it! Thanks!

KKimj:
Spec: https://mlir.llvm.org/docs/Dialects/Builtin/#sparseelementsattr

src/vcgen.cpp
```diff
@@ -670,6 +670,34 @@ optional<string> encodeOp(State &st, mlir::ConstantOp op) {
 
     st.regs.add(op, Integer(i.getSExtValue(), bw));
     return {};
+  } else if (mlir::SparseElementsAttr sparseAttr = attr.dyn_cast<mlir::SparseElementsAttr>()) {
+    std::vector<z3::expr> sparse_values;
```
makesource, line 674: I think it's good to follow the same c++ naming convention as MLIR code. (i.e use camelcase rather than snake case)

KKimj: Okay~

KKimj: Please check!
I resolved it.
Thanks!

PR #93, opened 2021-08-12
Implement more solver-independent wrappers

strikef:
This PR implements two new wrapper classes for solvers
- ~~`ExprVec` is a wrapper for vector of `Expr`s.~~ Removed
- `Context` is a wrapper for solver contexts.

Also, new semantics for `Expr` has been introduced
- `Expr`s cannot be constructed using constructors: instead, class static methods construct commonly used values.

Also, in order to reduce the mental load of tracking down aliases and mutations, `Expr` objects...
- are immutable: every methods create a new object instead of mutating itself
- ~~cannot be copy-assigned: one must use `clone()` method to do so, making sure that user acknowledges the object is being copied.~~ Excluded as of now
- ~~clearly distinguishes move and copy: this makes reasoning about aliases easier (*because there's virtually no alias*), reduces confusion, and helps avoid giant wave of compiler error messages regarding deleted operators.~~ Excluded as of now

### DIFF
--- a/src/smt.cpp
+++ b/src/smt.cpp
@@ -297,14 +297,6 @@ Expr ExprVec::fitsInDims(const ExprVec &sizes) const {
 }
 } // namespace smt
 
-
-void test() {
-  auto ct = smt::ContextBuilder().useZ3().build().value();
-  auto e1 = ct.bvVal(32, 32);
-  auto e2 = ct.bvVal(32, 32);
-  e1.add(e2);
-}
-
 llvm::raw_ostream& operator<<(llvm::raw_ostream& os, const smt::expr &e) {
   std::stringstream ss;
   ss << e;
--- a/src/smt.cpp
+++ b/src/smt.cpp
@@ -11,12 +11,18 @@ namespace {
 // Fn is simply declared because std::function with template arguments works
 // poorly. :(
 template<class T, class Fn>
-optional<T> fmap(const optional<T> &x, Fn fn) {
+optional<T> fmap(const optional<T> &x, Fn &&fn) {
   if (!x)
     return std::nullopt;
   return {fn(*x)};
 }
 
+template<class Fn>
+std::optional<z3::expr> fmap(z3::context *ctx, Fn &&fn) {
+  if (!ctx)
+    return std::nullopt;
+  return std::make_optional(fn(ctx));
+}
 }
 
 namespace smt {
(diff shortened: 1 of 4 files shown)


src/smt.cpp
```diff
@@ -105,21 +112,199 @@ string or_omit(const expr &e) {
...
+
+  return Expr(this->ctx, std::move(z3_expr));
 }
 
+ExprVec Expr::toNDIndices(const ExprVec &dims) const {
+  assert(dims.exprs.size() > 0);
+
+  auto idx_1d = this->clone();
+  auto expanded_exprs = ExprVec::withCapacity(this->ctx, dims.exprs.size());
+  std::for_each(dims.exprs.crbegin(), dims.exprs.crend(),
```
strikef, line 184: Changed from `std::accumulate` to `std::for_each`, as both `Expr` and `ExprVec` aren't very copy-friendly (`std::accumulate` [requires that the acc object is CopyConstructible](https://en.cppreference.com/w/cpp/algorithm/accumulate))

src/smt.cpp
```diff
@@ -105,21 +112,199 @@ string or_omit(const expr &e) {
...
+  return Expr(this->ctx, std::move(z3_expr));
+}
+
+Expr Expr::udiv(const Expr& rhs) const {
+  auto z3_expr = fmap(this->z3_expr, [&rhs](auto e) { return z3::udiv(e, *rhs.z3_expr); });
+  
+  return Expr(this->ctx, std::move(z3_expr));
+}
+
+Expr Expr::add(const Expr& rhs) const {
```
strikef, line 205: Maybe in the future...
```c++
auto z3_expr = fmap(this->z3_expr, ...);
auto cvc5_expr = fmap(this->cvc5_expr, ...);
...
return Expr(this->ctx, std::move(z3_expr), std::move(cvc5_expr), ...);
```

src/smt.cpp
```diff
@@ -105,21 +112,199 @@ string or_omit(const expr &e) {
...
+  return Expr(this->ctx, std::move(z3_expr));
+}
+
+Expr Expr::ugt(const Expr& rhs) const {
+  auto z3_expr = fmap(this->z3_expr, [&rhs](auto e) { return z3::ugt(e, *rhs.z3_expr); });
+  
+  return Expr(this->ctx, std::move(z3_expr));
+}
+
+Expr Expr::boolAnd(const Expr& rhs) const {
```
strikef, line 235: I also thought of naming it to `lAnd()`, but coming up with 'logical and' from 'lAnd' seems next to impossible :o
`logicalAnd()` or `logicAnd()` might work, but both are somewhat verbose...

src/smt.cpp
```diff
@@ -105,19 +112,189 @@ string or_omit(const expr &e) {
...
+  for (size_t i = 1; i < this->exprs.size(); i++) {
+    idx = idx.mul(dims.exprs[i]).add(this->exprs[i]);
+  }
+  return idx;
+}
+
+Expr ExprVec::fitsInDims(const ExprVec &sizes) const {
+  assert(this->exprs.size() == sizes.exprs.size());
+
+  Expr cond = this->ctx->boolVal(true);
```
strikef, line 292: An `Expr` or `ExprVec` should carry a `Context*` to handle cases like this.

src/smt.cpp
```diff
@@ -105,19 +112,180 @@ string or_omit(const expr &e) {
...
+  }
+}
+
+Expr Context::bvVal(const uint32_t val, const size_t sz) {
+  auto z3_expr = fmap(z3_ctx, [val, sz](auto ctx){ return ctx->bv_val(val, sz); });
+
+  return Expr(this, std::move(z3_expr));
+}
+
+Expr Context::bvConst(char* const name, const size_t sz) {
```
strikef, line 147: In long term, this should be a method of `Expr`

src/smt.cpp
strikef: The name can simply be `and`. However, it should internally disambiguate the bitwise/logical operation!

src/smt.h
```diff
@@ -26,19 +28,81 @@ expr fitsInDims(const std::vector<expr> &idxs,
...
-
-  Expr(std::optional<z3::expr> z3_expr): z3_expr(z3_expr) {}
+  
+  Expr(Context* const ctx) : ctx(ctx) {};
+  Expr(Context* const ctx, std::optional<z3::expr> &&z3_expr);
 
 public:
+  Expr(Expr&& from) = default;
+  Expr& operator=(Expr &&from) = default;
+  Expr clone() const;
```
strikef, line 67: Exclude move/clone semantics for now

src/smt.h
```diff
@@ -26,19 +28,81 @@ expr fitsInDims(const std::vector<expr> &idxs,
...
+
+public:
+  ExprVec(ExprVec&& from) = default;
+  ExprVec& operator=(ExprVec &&from) = default;
+  ExprVec clone() const;
+
+  ExprVec simplify() const;
+  Expr to1DIdx(const ExprVec &dims) const;
+  Expr fitsInDims(const ExprVec &sizes) const;
 };
```
strikef, line 104: Use std::vector<Expr> for now

strikef (commented):
Also, keep a static `Context` and hide it from outside!

strikef:
Resolved (hopefully all) issues raised.
More functions should be integrated into `Expr` function in the future, but now looks like a good point to check the progress.

One thing that bugs me is the `bool*` methods. Apparently `and` and `or` conflicts with [operator alternatives](https://en.cppreference.com/w/cpp/language/operator_alternative), so we have to come up with another name I guess

src/smt.cpp
aqjune: Hmm.. what about operator& ?

src/smt.cpp
```diff
@@ -173,19 +197,102 @@ string or_omit(const std::vector<expr> &evec) {
...
+}
+
+Expr Expr::boolOr(const Expr& rhs) const {
+  auto z3_expr = fmap(this->z3_expr, [&rhs](auto e) { return e || *rhs.z3_expr; });
+  
+  return Expr(std::move(z3_expr));
+}
+
+Expr Expr::bvVal(const uint32_t val, const size_t sz) {
+  auto z3_expr = sctx.z3Map([val, sz](auto &ctx){ return ctx.bv_val(val, sz); });
```
aqjune, line 280: Would fmap work?

src/smt.h
```diff
@@ -48,19 +46,34 @@ sort bvSort(unsigned bw);
 sort boolSort();
 sort arraySort(const sort &domain, const sort &range);
 
+class Expr;
+using ExprVec = std::vector<Expr>;
```
aqjune, line 50: Let's remove this and use std::vector<Expr>. :)

src/smt.cpp
```diff
@@ -14,7 +14,32 @@ z3::expr_vector toExprVector(const vector<smt::expr> &vec) {
 }
 
 namespace smt {
+class Context {
+
+public:
+  std::optional<z3::context> z3_ctx;
```
aqjune, line 20: No std:: in .cpp

aqjune:
Looks good with minor comments! Let's edit the description of this PR as well.

src/smt.cpp
strikef: We implemented every arithmetic operators for `Expr` as methods, so I'm kinda concerned that it may break the consistency...

src/smt.cpp
strikef: That also works, if we define another fmap.
```c++
template <class T, class Fn>
auto fmap(optional<T> &x, Fn &&fn) {
...
}
```

Note that this one does not have **const** in the parameter type.

src/smt.cpp
aqjune: operator& can be also declared as a member function, so it is fine.

src/smt.cpp
strikef: Should we remove *all* `std::` in the .cpp file? (Like move, reverse, nullopt, ...)

src/smt.h
strikef: Fixed

src/smt.cpp
```diff
@@ -10,12 +11,18 @@ namespace {
 // Fn is simply declared because std::function with template arguments works
 // poorly. :(
 template<class T, class Fn>
-optional<T> fmap(const optional<T> &x, Fn fn) {
+optional<T> fmap(const optional<T> &x, Fn &&fn) {
```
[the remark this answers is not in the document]
strikef: Accepting Fn as rvalue doesn't actually 'move' the functions, but rather allows the use of 'move-only capture variables` according to [this](http://www.cplusplus.com/forum/general/202662/) thread.

src/smt.cpp
strikef: Implemented as `fupdate()`

PR #98, opened 2021-08-17
Add inverseMappings to MemRef Layout encoding

makesource:
Before we support various memref operations, we need the definition of inverseMapping of layout map.
This is because many arithmetic operations like conv, add, mul can be encoded more easily and efficiently if we abstract concrete MemRef's layout.
In other words, it is much simpler to find the expression for %memref[i][j] and then encode it in memory by giving concrete layout information.

In this PR, we changed to use lambda when encoding layout to better express that is a function.
And add new inverseMappings which represent inverse function of layout mapping.
These will be used later when we support additional memref operations.

Since there is no operation that uses inverseMappings yet, nothing changed from current smt encoding.
I also checked `ctest` works well without problems.

### DIFF
--- a/src/smt.h
+++ b/src/smt.h
@@ -20,7 +20,6 @@ std::vector<expr> simplifyList(const std::vector<expr> &exprs);
 
 expr to1DIdx(const std::vector<expr> &idxs,
                 const std::vector<expr> &dims);
-expr to1DIdxWithLayout(const std::vector<expr> &idxs, expr layout);
 expr fitsInDims(const std::vector<expr> &idxs,
                 const std::vector<expr> &sizes);
 
--- a/src/smt.h
+++ b/src/smt.h
@@ -30,7 +30,12 @@ std::string or_omit(const std::vector<expr> &evec);
 // TODO: these functions must be member functions of Expr
 expr substitute(expr e, const std::vector<expr> &vars,
                 const std::vector<expr> &values);
+expr implies(const expr &a, const expr &b);
 expr forall(const std::vector<expr> &vars, const expr &e);
+expr lambda(const expr &var, const expr &e);
+expr lambda(const std::vector<expr> &vars, const expr &e);
+expr select(const expr &arr, const expr &idx);
+expr select(const expr &arr, const std::vector<expr> &idxs);
 expr mkFreshVar(const sort &s, std::string &&prefix);
 expr mkVar(const sort &s, std::string &&name);
 expr mkBV(uint64_t i, unsigned bw);
--- a/src/smt.h
+++ b/src/smt.h
@@ -27,6 +27,7 @@ std::string or_omit(const expr &e);
 std::string or_omit(const std::vector<expr> &evec);
 
 // TODO: these functions must be member functions of Expr
+expr init();
 expr substitute(expr e, const std::vector<expr> &vars,
                 const std::vector<expr> &values);
 expr implies(const expr &a, const expr &b);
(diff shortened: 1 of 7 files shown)


src/smt.h
```diff
@@ -20,24 +20,30 @@ std::vector<expr> simplifyList(const std::vector<expr> &exprs);
...
+expr select(const expr &arr, const expr &idx);
+expr select(const expr &arr, const std::vector<expr> &idxs);
 expr mkFreshVar(const sort &s, std::string &&prefix);
 expr mkVar(const sort &s, std::string &&name);
 expr mkBV(uint64_t i, unsigned bw);
 expr mkBool(bool b);
 func_decl mkUF(const sort &domain, const sort &range, std::string &&name);
 func_decl mkUF(const std::vector<sort> &domain, const sort &range,
                std::string &&name);
+expr fapply(const func_decl &func, const std::vector<expr> &vars);
```
makesource, line 46: I name it `fapply` which means "function apply".

src/value.h
```diff
@@ -174,19 +174,42 @@ class MemRef {
...
-      const smt::expr &inbounds):
-      indVars(indVars), expr(expr), inbounds(inbounds) {}
-
-    Layout eval(smt::model mdl) const {
-      return { indVars, mdl.eval(expr).simplify(),
-               mdl.eval(inbounds).simplify() };
-    }
+        const smt::expr &layout,
+        const smt::expr &inbounds,
+        bool useUF = false); // encode "mapping" using uninterpreted function
```
makesource, line 205: Here we give `useUF` options to select mapping encoding.
This may not work well in concrete cases like `subview`, but work faster in very simple store, load operations.
I left this options for further use in performance testing.

src/smt.h
makesource: Also to avoid direct use of `toExprVector`

src/value.cpp
```diff
@@ -509,6 +493,68 @@ expr Tensor::to1DArrayWithOfs(
...
+
+MemRef::Layout::Layout(const std::vector<smt::expr> &indVars,
+    const smt::expr &layout,
+    const smt::expr &inbounds,
+    bool useUF): indVars(indVars), inbounds(init()), mapping(init()), precondition(init()) {
+  if (useUF) {
+    vector<smt::sort> domains(indVars.size(), Index::sort());
+    auto layoutFn = mkUF(domains, Index::sort(), freshName("layoutFn"));
+    auto layoutFnExpr = fapply(layoutFn, indVars);
+    expr condition = (layoutFnExpr == layout);
```
makesource, line 526: give concrete semantic of uninterpreted mapping function.

src/value.cpp
```diff
@@ -509,6 +493,68 @@ expr Tensor::to1DArrayWithOfs(
...
+      condition = condition && smt::select(inverse, layoutFnExpr) == indVars[i];
+    }
+    this->inbounds = smt::lambda(indVars, inbounds);
+    this->mapping = smt::lambda(indVars, layoutFnExpr);
+    this->inverseMappings = inverseMappings;
+    this->precondition = smt::forall(indVars, smt::implies(inbounds, condition));
+  } else {
+    expr condition = mkBool(true);
+    vector<expr> inverseMappings;
+    for (unsigned i = 0; i < indVars.size(); i ++) {
```
aqjune, line 543: Can we merge this duplicated part in if and else?

makesource: This is intended to make it easier to remove certain encodings later.

PR #105, opened 2021-08-18
More functionalities for `smt::Expr`

strikef:
This PR implements `substitute()`, `implies()`, and `select()` for `Expr` class.

I'm also planning on implementing `lambda()` and `forall()` for `Expr`, but the problem with these functions are that they accept a vector of `Expr`s as their first parameter. It doesn't look natural to implement them as `Expr` method...
I'll fuily open this PR once I get to implement these two in a nice form.

### DIFF
--- a/src/smt.cpp
+++ b/src/smt.cpp
@@ -216,6 +216,38 @@ Expr Expr::simplify() const {
   return Expr(move(z3_expr));
 }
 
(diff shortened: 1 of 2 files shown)


strikef:
The functions implemented here will be integrated into `add-cvc5` branch.

PR #106, opened 2021-08-18
Replace `z3::expr` in `Index` with `smt::Expr`

strikef:
This PR 'tries' replacing the `z3::expr` in `Index` with `smt::Expr` into `Index`. (Both of them exist as of now)

`Index` has some methods that depends on `z3::expr`. `smt::Expr` cannot directly replace the `z3::expr`, so these methods are marked as 'deprecated'. They will be removed once we actually start replacing the exprs.

After merging this PR, I'll start replacing the exprs in `Float` and `Integer` as well.

### DIFF
--- a/src/smt.cpp
+++ b/src/smt.cpp
@@ -303,6 +303,22 @@ Expr Expr::operator|(const Expr &rhs) {
   return Expr(move(z3_expr));
 }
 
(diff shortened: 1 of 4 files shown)


strikef:
The `expr`s will be replaced all at once when the `smt::Expr` is completely implemented.
Closing this PR for now.

PR #111, opened 2021-08-25
Make Tensor, MemRef class inherit Shaped class

makesource:
In this PR, we make a common class of Tensor, MemRef class which contains common index (dimension) based logic.
Maybe linalg operations related with both Tensor, MemRef are fit here.
These operations will be added in following PR.

### DIFF
--- a/src/encode.cpp
+++ b/src/encode.cpp
@@ -328,7 +328,7 @@ optional<string> encodeOp(State &st, mlir::memref::LoadOp op) {
   for (auto idx0: op.indices())
     indices.emplace_back(st.regs.get<Index>(idx0));
 
-  auto [Expr, success] = m.load(indices);
+  auto [Expr, success] = m.get(indices);
   if (auto vt = fromExpr(move(Expr), op.getType())) {
     st.regs.add(op, move(*vt));
     st.wellDefined(op.getOperation(), move(success));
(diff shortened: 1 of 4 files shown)


src/encode.cpp
```diff
@@ -412,8 +410,9 @@ optional<string> encodeOp(State &st, mlir::memref::BufferCastOp op) {
 
   } else {
     vector<Expr> idxs = createBoundIndexVars(memrefTy.getRank());
-    auto tVal = tensor.get(idxs);
-    auto [mVal, success] = memref.load(idxs);
+    auto [tVal, tSuccess] = tensor.get(idxs);
+    auto [mVal, mSuccess] = memref.get(idxs);
+    auto success = tSuccess & mSuccess;
```
aqjune, line 415: Is there a case where tSuccess != mSuccess?

src/value.h
```diff
@@ -88,7 +88,16 @@ class Integer {
   Integer eval(smt::Model m) const;
 };
 
-class Tensor {
+class Shaped {
```
aqjune, line 91: Let's rename this to `ShapedValue`

src/encode.cpp
makesource: I think `mSuccess` is more strict than `tSuccess`.
```
mSuccess = dimension inbounds && bid, offset inbounds
tSuccess = dimension inbounds
```
Here buffer_cast op only accept when tensor, memref dimensions are match.
So the first term will be same in both success.

PR #114, opened 2021-08-26
Support linalg::conv encoding

makesource:
In this PR, we added `linalg::conv`operations encoding.
Main convolution calculation is done by `ShapedValue::conv`, and memref's conv operation just store result values into Memory.
For simplicity, currently we support only plain layout mapped MemRef.
This constraint may be relaxed later..

### DIFF
--- a/src/value.cpp
+++ b/src/value.cpp
@@ -722,11 +722,11 @@ Expr MemRef::conv(const MemRef &input,
     const MemRef &filter,
     const std::vector<smt::Expr> strides,
     const std::vector<smt::Expr> dilations) {
-  auto [indices, expr] = ((ShapedValue *) &input)->conv(filter, strides, dilations);
+  auto [indices, expr] = input.ShapedValue::conv(filter, strides, dilations);
 
   // we splat results into 1D memory layout
   auto idx = Index::var("outputIdx", VarType::BOUND);
-  auto outputIndices = getInverseIndices(idx);
+  auto outputIndices = layout.getInverseIndices(idx);
   auto outputExpr = expr.substitute(indices, outputIndices);
   auto outputArray = Expr::mkLambda(idx, outputExpr);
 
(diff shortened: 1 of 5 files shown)


aqjune:
I implemented MemRef alloc via d10b830a4aa8fba2fe924545273871eb82b00c6e.
Instead of MemRef copy, I implemented tensor_store (41ab9ba91118b69aed9b572be66ca42c7f503d71).
I guess this can be used to write tests for conv(tensor) <-> conv(memref) as well?
Also, MemRef copy is tensor_load + tensor_store, so its naive version might be straightforward to implement; a question is whether it can be encoded more efficiently than this.

src/value.cpp
```diff
@@ -678,7 +678,7 @@ Expr MemRef::storeArray(
 Expr MemRef::isInBounds() const {
   auto numelem = m->getNumElementsOfMemBlock(bid);
   auto memrefSize = get1DSize();
-  return numelem.uge(memrefSize) & ((Expr)offset).ult(numelem - memrefSize);
+  return numelem.uge(memrefSize) & ((Expr)offset).ule(numelem - memrefSize);
```
makesource, line 681: This might be undiscovered bugs in `isInBounds` encoding.
Offset can be zero if `numelem` and `memrefSize` has same value.. but in previous `ult` encoding this was not allowed.

src/value.cpp
```diff
@@ -718,6 +718,25 @@ MemRef MemRef::subview(const vector<Expr> &offsets,
...
+
+  // we splat results into 1D memory layout
+  auto idx = Index::var("outputIdx", VarType::BOUND);
+  auto outputIndices = layout.getInverseIndices(idx);
+  auto outputExpr = expr.substitute(indices, outputIndices);
+  auto outputArray = Expr::mkLambda(idx, outputExpr);
+
+  // store output memref
+  auto success = isInBounds() & input.isInBounds() & filter.isInBounds() &
+    storeArray(outputArray, Index::zero(), get1DSize());
```
makesource, line 735: check whether "output", "input", "filter" MemRef is inbounds.

src/value.cpp
```diff
@@ -718,6 +736,25 @@ MemRef MemRef::subview(const vector<Expr> &offsets,
...
+
+  // we splat results into 1D memory layout
+  auto idx = Index::var("outputIdx", VarType::BOUND);
+  auto outputIndices = layout.getInverseIndices(idx);
+  auto outputExpr = expr.substitute(indices, outputIndices);
+  auto outputArray = Expr::mkLambda(idx, outputExpr);
+
+  // store output memref
+  auto success = isInBounds() & input.isInBounds() & filter.isInBounds() &
+    noalias(input) & noalias(filter) & storeArray(outputArray, Index::zero(), get1DSize());
```
makesource, line 753: Check whether "output", "input", "filter" MemRef is inbounds.
Also I added noalias condition for output MemRef

makesource:
@aqjune 
I added litmus test for conv(MemRef) <-> conv(Tensor).
See `tests/litmus/linalg-ops/convolution.src.mlir`, `tests/litmus/linalg-ops/convolution.tgt.mlir`!

PR #126, opened 2021-09-09
Implement CI base image daily update

strikef:
This PR implements a GitHub action script to update the CI base image every day.
- Fetches LLVM, Z3, CVC5 repository every day, update if necessary, and construct Docker image with them
- Upload the image to DockerHub so that they're available for each CI runs.

Also, this is the first attempt to create a PR from a branch in the main repo, which allows the validation through CI.
From now on, we should directly push our commits into a branch in the upstream (which is this repo). And we should make PRs only from branches in the main repo.


strikef:
Just realized that the forks of this repo will also run this CI every day, which is seriously redundant :(

strikef:
Fortunately there's a way to limit the execution of this script to a specific repository. This way the forked repositories won't try to run this script, unless the fork owners edit the script ;)

PR #128, opened 2021-09-13
Implement option to fold fadd and fmul identity

strikef:
This PR implements a set of new behaviors related to identity to Float::add() and Float::mul().

- When `--add-identity` flag is set, Float(a).add(Float(0.0)) directly returns Float(a) and Float(0.0).add(Float(a)) directly returns Float(a).
- When `--mul-identity` flag is set, Float(a).mul(Float(1.0)) directly returns Float(a) and Float(1.0).mul(Float(a)) directly returns Float(a).
- Litmus tests are added to show the changed behavior

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1,6 +1,7 @@
 #include "memory.h"
 #include "smt.h"
 #include "vcgen.h"
+#include "value.h"
 #include "llvm/Support/CommandLine.h"
 #include "llvm/Support/Debug.h"
 #include "llvm/Support/PrettyStackTrace.h"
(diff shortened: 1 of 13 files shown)


strikef:
This PR includes the bugfix in #127 (the bug was found while implementing this feature)

src/main.cpp
```diff
@@ -53,6 +54,16 @@ llvm::cl::opt<bool> arg_associative_sum("associative",
...
+llvm::cl::opt<bool> arg_add_identity("add-identity",
+  llvm::cl::desc("Assume that fpadd 0.0 is an identity operation "
+                 "(experimental)"),
+  llvm::cl::init(false));
+
+llvm::cl::opt<bool> arg_mul_identity("mul-identity",
+  llvm::cl::desc("Assume that fpmul 1.0 is an identity operation "
+                 "(experimental)"),
+  llvm::cl::init(false));
+
```
strikef, line 66: Seongwon suggested that since these are 'safe' operations, we can use this 'rules' as an default behavior, not as options.
I think it makes sense, too. But for now I made them optional just in case.

src/smt.cpp
```diff
@@ -644,16 +644,21 @@ Expr Expr::substitute(
...
+      res = is_or ? (res || temp) : (res && temp);
+    })
+  
   IF_CVC5_ENABLED(
-    temp = cvc5 && cvc5->getId() == e2.cvc5->getId();
-    res = is_or ? (res || temp) : (res && temp));
+    if (cvc5) {
+      temp = cvc5->getId() == e2.cvc5->getId();
+      res = is_or ? (res || temp) : (res && temp);
+    })
```
strikef, line 658: Duplicate of #127

strikef:
Will re-implement this using SMT precondition

PR #131, opened 2021-09-17
Encode associativeSum using uninterpreted hash function

makesource:
Encode associativeSum expression by using uninterpreted function.
The mlir-tv checks only negated refinement, there's no need for qualifiers in hash function.


makesource:
This implementation has some problems :(
I'll reopen pull request after resolving issues.

PR #134, opened 2021-09-24
Encode `sum_fn`'s associativity using uninterpreted hash function

makesource:
Encode `sum_fn`'s associativity by using uninterpreted function.
`sum_fn`'s equality is given by preconditions which are built when encoding src, tgt smt arrays.

We only add equality conditions when two smt arrays have same length to reduce the number of candidate pairs.
For each candidate pairs, we use distinct `hashfn`s to check whether they can be hashed into different values.

### DIFF
--- a/src/abstractops.cpp
+++ b/src/abstractops.cpp
@@ -74,7 +74,7 @@ Expr mkZeroElemFromArr(const Expr &arr) {
   return Expr::mkBV(0, bvsz);
 }
 
-optional<FnDecl> sumfn, dotfn, fpaddfn, fpmulfn;
+optional<FnDecl> sumfn, dotfn, fpaddfn, fpmulfn, hashfn;
 
 
 Expr fpAdd(const Expr &f1, const Expr &f2) {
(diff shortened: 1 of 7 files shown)


src/abstractops.cpp
```diff
@@ -159,12 +156,36 @@ Expr dot(const Expr &a, const Expr &b, const Expr &n) {
...
 
+Expr getAssociativePrecondition() {
+  Expr precond = Expr::mkBool(true);
+  for (unsigned i = 0; i < staticArrays.size(); i ++) {
+    for (unsigned j = i + 1; j < staticArrays.size(); j ++) {
+      auto [a, an, asum] = staticArrays[i];
+      auto [b, bn, bsum] = staticArrays[j];
+      uint64_t alen, blen;
+      if (!an.isUInt(alen) || !bn.isUInt(blen) || alen != blen) continue;
+      FnDecl hashfn(Float::sort(), Index::sort(), freshName("hash"));
```
makesource, line 172: Use a unique hash for each pair of candidates

makesource:
I'll reopen this PR through upstream branch.

PR #137, opened 2021-09-27
Implement tensor.extract_slice

inwchun:
implemented to support extract slice


inwchun:
currently fixing code since it has an infinite loop in mlir-tv-testsuite.

inwchun:
resolved

tests/litmus/tensor-ops/extract1d.tgt.mlir
```diff
@@ -0,0 +1,8 @@
+// VERIFY
```
aqjune, line 1: // VERIFY is not needed in tgt

src/encode.cpp
```diff
@@ -381,6 +381,70 @@ optional<string> encodeOp(State &st, mlir::tensor::FromElementsOp op) {
...
+      strides.push_back(st.regs.get<Index>(op.strides()[varIdx++]));
+    else
+      strides.push_back(Index(s));
+  }
+
+  vector<smt::Expr> dims; 
+
+  unsigned j=0;
+  for (unsigned i = 0; i < resType.getRank(); i++) {
+    while((sizes[j] == Index(1)).isTrue()) {
```
aqjune, line 418: I guess there is `sizes[j].isUInt` or something like that

src/encode.cpp
```diff
@@ -381,6 +381,70 @@ optional<string> encodeOp(State &st, mlir::tensor::FromElementsOp op) {
...
+optional<string> encodeOp(State &st, mlir::tensor::ExtractSliceOp op) {
+  vector<Expr> offsets, sizes, strides;
+  auto src = st.regs.get<Tensor>(op.getOperand(0));
+  auto srcType = op.getOperand(0).getType().dyn_cast<mlir::ShapedType>();
+  auto res = op.getResult();
+  auto resType = res.getType().dyn_cast<mlir::ShapedType>();
+
+  auto varIdx = 0;
+  for (auto s: extractFromI64ArrayAttr(op.static_offsets())){
+    if(s ==-0x8000000000000000)
```
aqjune, line 394: a space after `if`
BTW, where did you find this check (a comparison with `-0x8000000000000000`)?

src/encode.cpp
```diff
@@ -381,6 +381,70 @@ optional<string> encodeOp(State &st, mlir::tensor::FromElementsOp op) {
...
+      offsets.push_back(st.regs.get<Index>(op.offsets()[varIdx++]));
+    else
+      offsets.push_back(Index(s));
+  }
```
aqjune, line 398: Can we define a lambda that has this pattern?
Then, we can simply replace these three loops with three lambda calls.

src/encode.cpp
```diff
@@ -381,6 +381,70 @@ optional<string> encodeOp(State &st, mlir::tensor::FromElementsOp op) {
...
+  vector<smt::Expr> dims; 
+
+  unsigned j=0;
+  for (unsigned i = 0; i < resType.getRank(); i++) {
+    while((sizes[j] == Index(1)).isTrue()) {
+      j++;
+    }
+
+    // check if output tensor matches size
+    assert((sizes[j] == Index(resType.getDimSize(i))).isTrue());
```
aqjune, line 423: Is this still true if the output dimension is not constant?

aqjune:
Please add a test that has `tensor.extract_slice %v[%arg0,%arg1][..][..]: .. to tensor<?x?xf32>`.
A tensor type may have an unknown size (`?`).

inwchun:
resolved all reviews

src/encode.cpp
inwchun: found a simple function for reading inputs

src/encode.cpp
inwchun: added a case when dimension size is unknown

PR #142, opened 2021-09-30
Resolve compiler warnings

strikef:
- Resolved unused variable (`Wunused-variable`) warnings
- Resolved control flow reaches end of non-void function (`Wreturn-type`) warnings


strikef:
It seems that these variables are necessary in case we're linking against single solver :o

aqjune:
The variables are needed because they are used in assertions. :)

PR #145, opened 2021-10-01
Change `Float` encoding

strikef:
This PR introduce a new way of encoding `Float`s in `mlir-tv`.

`Float` is encoded as BitVector`[3+VALUE_BITS]`.
- The MSB represents the sign (0: positive, 1: negative)
- The following two bits represent type of the value (00: Finite value, 01: Inf, 10: NaN)
- `VALUE_BITS` is equal to `FP_BITS` before this PR.

### DIFF
--- a/src/abstractops.cpp
+++ b/src/abstractops.cpp
@@ -12,14 +12,25 @@ static string freshName(string prefix) {
 }
 
 namespace {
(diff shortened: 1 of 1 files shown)


aqjune:
Rebase please~

PR #125, opened 2021-09-08
Implement experimental associative sum encoding

makesource:

PR #147, opened 2021-10-01
Revert multiset encoding for benchmark

makesource:
Revert previous multi sets encoding in https://github.com/aqjune/mlir-tv/pull/125.
But in previous PR, bag equality with uninterpreted function is not well supported in CVC5..
To resolve this issue, I added bag equality precondition in `getAssociativePrecondition()` and this works well!

cf. Currently CVC5 support ackermannization in very limited condition. So I just encoded it in precondition :(

### DIFF
--- a/src/abstractops.cpp
+++ b/src/abstractops.cpp
@@ -213,11 +213,11 @@ Expr getAssociativePrecondition() {
         FnDecl hashfn(Float::sort(), Index::sort(), freshName("hash"));
 
         auto aVal = hashfn.apply(a.select(Index(0)));
-        for (unsigned i = 1; i < alen; i ++)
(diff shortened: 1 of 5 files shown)


aqjune:
Is there any reason why `associativeSum` is separately defined?
`sum` is also updating `staticArrays` and preconditions for those are being encoded. Therefore `sum` is also associative.

makesource:
Although `sum` is also updating staticArrays.. but this preconditions will only be used when `--associative` flag is given.
Without this flag, `sum` is just an uninterpreted function and is not associative.
Well.. I think it would be better to rename `associativeSum` to `multisetSum`.

makesource:
I'll reopen this PR with upstream branch :)

PR #149, opened 2021-10-04
Implement broadcast feature for tosa.add

inwchun:
implemented broadcast feature for tosa.add


aqjune:
Q: Where did you get the broadcasting algorithm? I couldn't find how it works from the MLIR document.

inwchun:
I couldn't find the document for it, but I looked up several test cases in the testsuite and llvm-project.

src/encode.cpp
```diff
@@ -117,6 +117,63 @@ static mlir::Type getTensorElemTy(mlir::Value v) {
...
+  auto resRank = max(ty0.getRank(), ty1.getRank());
+  auto inVars0 = Index::boundIndexVars(resRank);
+  auto inVars1 = Index::boundIndexVars(resRank);
+
+  vector<Expr> outVars0, outVars1, resDims;
+  for(int i = 0; i < resRank; i++) {
+    auto idx0 = ty0.getRank() - 1 - i;
+    auto idx1 = ty1.getRank() - 1 - i;
+
+    if(idx0 >= 0 && idx1 >= 0) {
```
aqjune, line 134: idx0 and idx1 are unsigned integers..?

src/encode.cpp
```diff
@@ -117,6 +117,63 @@ static mlir::Type getTensorElemTy(mlir::Value v) {
...
+      auto d1 = ty0.getDimSize(idx0);
+      auto d2 = ty1.getDimSize(idx1);
+      assert(d1 == 1 || d2 == 1 || d1 == d2);
```
aqjune, line 137: What if `ty0.getDimSize(idx0)` has an unknown size?

src/encode.cpp
```diff
@@ -117,6 +117,63 @@ static mlir::Type getTensorElemTy(mlir::Value v) {
...
+        outVars0.insert(outVars0.begin(), Index(0));
+      else 
+        outVars0.insert(outVars0.begin(), inVars0[idx0]);
+
+      if(d2 == 1)
+        outVars1.insert(outVars1.begin(), Index(0));
+      else 
+        outVars1.insert(outVars1.begin(), inVars1[idx1]);
+    }
+    else {
```
aqjune, line 150: `} else {`

tests/litmus/tosa-ops/add_broadcast5.src.mlir
```diff
@@ -0,0 +1,7 @@
+// VERIFY
+
+func @add(%arg0: tensor<i32>, %arg1: tensor<17x16x15x14xi32>) -> tensor<17x16x15x14xi32> {
+  //  CHECK-NEXT: reshape
```
aqjune, line 4: CHECK-NEXT?

src/encode.cpp
inwchun: getRank returns int64_t, so I think it would be int64_t, but I'll cast it with int64_t

tests/litmus/tosa-ops/add_broadcast1.tgt.mlir
```diff
@@ -0,0 +1,4 @@
+func @add(%arg0: tensor<1xf32>, %arg1: tensor<10x9x8x7xf32>) -> tensor<10x9x8x7xf32> {
+  %0 = "tosa.add"(%arg0, %arg1) : (tensor<1xf32>, tensor<10x9x8x7xf32>) -> tensor<10x9x8x7xf32>
```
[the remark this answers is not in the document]
inwchun: tests validate whether the broadcasting is correctly implemented for each case (whether 1xf32 exanded 10x9x8x7xf32 in this case)

src/encode.cpp
inwchun: hmm... I've searched the test suite but I can't find tosa.add with ? dimension. fixed to return the original tensor if there is an unknown dim.

tests/litmus/tosa-ops/add_broadcast1.tgt.mlir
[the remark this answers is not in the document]
aqjune: But the pair can still pass if `tosa.add` is simply defined to always yield a zero tensor...
What about doing `tensor.extract` to check whether broadcasting is done successfully?

src/encode.cpp
```diff
@@ -133,8 +194,9 @@ encodeBinaryOp(State &st, OpTy op, mlir::Value arg0, mlir::Value arg1,
     if (!elemty.isIntOrFloat())
       return "Unsupported element type";
 
-    auto a = st.regs.get<Tensor>(arg0);
-    auto b = st.regs.get<Tensor>(arg1);
+    auto tPair = getBroadcastTensor(st, arg0, arg1);
+    auto a = tPair.first;
+    auto b = tPair.second;
```
aqjune, line 199: Would
```
auto [a, b] = getBroadcastTensor(st, arg0, arg1);
```
work?

PR #152, opened 2021-10-05
Implement tensor.insert_slice operation

inwchun:
Implementation of tensor.insert_slice


inwchun:
Adding test cases

inwchun:
added test cases

src/encode.cpp
```diff
@@ -435,6 +435,64 @@ optional<string> encodeOp(State &st, mlir::tensor::ExtractSliceOp op) {
...
+
+  Expr output = Expr::mkIte(((inIdxs[0] - offsets[0]) % strides[0]).isZero() &
+                          ((inIdxs[0] - offsets[0]) / strides[0]).ult(sizes[0]),
+                          src1.get(src1Idxs).first, src2Mapping);
+
+  for (unsigned i = 1; i < resType.getRank(); i++) {
+    output = Expr::mkIte(((inIdxs[i] - offsets[i]) % strides[i]).isZero() &
+                          ((inIdxs[i] - offsets[i]) / strides[i]).ult(sizes[i]),
+                          output, src2Mapping);
+  }
```
aqjune, line 490: What about representing the output as

`ite (the big conjunction, src1.get(src1Idxs).first, src2Mapping)`

rather than

`ite(ite(cond, ite(cond, src1.get(src1Idxs).first, src2Mapping), src2Mapping), src2Mapping)`?

aqjune:
Add a test with tensors of unknown sizes?

inwchun:
Resolved.

aqjune:
I updated the patch as following:
- Rely on UB to do in-bounds checking
- Unsupport the input if src1.rank != src2.rank (I found that `mlir/test/IR/core-ops.mlir` has such a case, but it's not clear to me how axis must be aligned...)
- Remove operator/ because the name does not clearly exhibit whether it is signed or not. Let's use udiv() instead.

PR #182, opened 2021-11-10
Implement more precise fp counting

strikef:
This PR implements precision-sensitive FP counting.
- F32 and F64 constants/variables are counted separately
- Each F32 and F64 constant found is also added respectively to F64 and F32 constant list, with possible rounding if necessary.


makesource (approved):
LGTM 👍

strikef:
I'll re-open this PR after implementing discrete variable counting as well

PR #184, opened 2021-11-13
Implement linalg.conv_2d_nchw_fchw

inwchun:
implementation of linalg.conv_2d_nchw_fchw


inwchun:
The result is correct, but timeout occurs (takes around 17000ms).

PR #185, opened 2021-11-14
New FP encoding for extension

strikef:


src/abstractops.cpp
```diff
@@ -90,12 +110,17 @@ AbsFpEncoding &getFpEncoding(mlir::Type ty) {
...
+      AbsFpEncoding* larger_fpty_enc, std::string &&fn_suffix)
      :semantics(semantics), fn_suffix(move(fn_suffix)) {
   assert(valuebits > 0);
-  value_bv_bits = valuebits;
+  // BWs for casting
+  limit_bv_bits = limitbits;
+  prec_bv_bits = precbits;
+  this->larger_fpty_enc = larger_fpty_enc;
+
+  value_bv_bits = limitbits + precbits + valuebits;
```
aqjune, line 123: What about keeping
`value_bv_bits = valuebits;`
and using `limitbits`/`precbits` in other places differently?

aqjune: I think I understood what it means.. nevermind

PR #188, opened 2021-11-16
enable multiple return for linalg.generic

inwchun:
enable multiple return for linalg.generic


inwchun:
VERIFY-INCORRECT test failing for pad_tensor & generate if vector input is given. (can't find reason)

PR #191, opened 2021-11-17
Improve abstract representation of double by adding limit and prec bits

strikef:
This PR implements const-to-absrepr mapping based on constant analysis.
- Abstract representation of constants now *almost* correctly reflect the actual ext/trunc relationship
- Assigning abstract representation to each constant is moved from `AbsFpEncoding::constant()` to `AbsFpEncoding::addConstants()`
- Abstract representation of each constant now also reflects the actual order of values
- `AbsFpEncoding::limit_bitwidth` and `AbsFpEncoding::prec_bitwidth` are determined based on constant analysis as well

**About *almost***: In order to reflect the order of values using the abstract representation, f64 constants are floored during the mapping process. 
**This is not what typical FP applications would expect (which would be rounding)**


src/abstractops.cpp
```diff
@@ -199,7 +203,57 @@ uint64_t AbsFpEncoding::getSignBit() const {
   return 1ull << value_bitwidth;
 }
 
-Expr AbsFpEncoding::constant(const llvm::APFloat &f) {
+void AbsFpEncoding::addConstantsFromAnalysis(const std::set<FPConstAnalysisResult>& analysis_set) {
```
aqjune, line 206: No std:: in .cpp

src/abstractops.cpp
```diff
@@ -199,7 +203,57 @@ uint64_t AbsFpEncoding::getSignBit() const {
...
+        // so assign different smaller_value
+        value_id += 1;
+      }
+      e_value = e_value.concat(e_prec);
+    } else {
+      // should not reach
+      llvm_unreachable("should not reach here!");
+    }
+
+    if (analysis.value.isNonZero()) {
```
aqjune, line 246: Let's hoist this if stmt to the top inside this loop:
```
if (analysis.value.isNonZero())
  continue;
```

strikef:
The [recent PR](https://github.com/aqjune/mlir-tv/pull/194) by @makesource fixes the issue where FP analysis missed out the constants inside linalg.generic block. I've checked by myself that our CI passes with both patches applied.

strikef:
It appears that limit_bitwidth and prec_bitwidth must be adjusted according to the analysis result. I'll re-open this PR once it is implemented.

strikef:
The above issue will be resolved in later PR

strikef:
The issue stated above is fixed: When constructing an AbsFpEncoding, FP const analysis results are used to set the minimum required limit/prec bits to correctly encode constants

PR #193, opened 2021-11-18
Analyze constant floating number in linalg.generic loop

makesource:
Analyzing process also consider linalg.generic inner loop.


makesource:
Oops.. I'll reopen in upstream branch :(

PR #194, opened 2021-11-18
Analyze constant floating number in linalg.generic loop

makesource:
Analyzing process also consider linalg.generic inner loop.


strikef:
This patch shows the expected (and correct) behavior when tested against my recent implementation.
Logic looks simple enough and seems valid, so LGTM :)

strikef (commented):
This patch shows the expected (and correct) behavior when tested against my recent implementation.
Logic looks simple enough and seems valid, so LGTM :)

PR #199, opened 2021-11-23
Implement value folding in `Expr::select()`

strikef:
This PR implements implicit value folding for `Expr::select()`
- Folding means that this operation will try to simplify certain smt expressions into a value
- If the array operand of `Expr::select()` is const_array, the selected element will be directly returned
- ~~Implemented `smtmatcher::Select` for expression analysis~~ This has been removed


strikef:
This PR is still in draft, because using the `Expr::select()` occasionally crashes with `Pure virtual function called!` error. It seems to have something to do with dimensions of the array, but as of now I'm still looking into it.

strikef:
Apparently the above issue was due to a bug in the previous version of Z3 :(

PR #201, opened 2021-11-25
Implement tosa.transpose

inwchun:
Implement 
tosa.transpose


inwchun:
tosa.transpose done

src/encode.cpp
```diff
@@ -957,6 +957,64 @@ void encodeOp(State &st, mlir::tosa::BitwiseXorOp op, bool) {
...
+  auto perms = st.regs.get<Tensor>(p);
+
+  vector<Expr> indVars = Index::boundIndexVars(input.getRank());
+  vector<Expr> dims, outVars;
+  vector<uint64_t> idxs;
+
+  for (unsigned i = 0; i < input.getRank(); i++) {
+    uint64_t v;
+    // We expect simplify() to succeed since perms is a small Tensor
+    assert(perms.get({Index(i)}).first.simplify().isUInt(v));
```
aqjune, line 986: Should we throw `UnsupportedException` rather than `assert`?

PR #205, opened 2021-11-30
Implement tosa.conv2d

inwchun:
Implement tosa.conv2d


inwchun:
all litmus tests succeed except conv2d2.mlir. This takes 360 seconds, but the result is correct (case of stride 2).

src/encode.cpp
```diff
@@ -970,6 +970,55 @@ void encodeOp(State &st, mlir::tosa::BitwiseXorOp op, bool) {
...
+                            srcDims[2] + pad[2] + pad[3], srcDims[3]};
+
+  auto cond = padInd[1].uge(pad[0]) & padInd[1].ult(pad[0] + srcDims[1]) &
+                padInd[2].uge(pad[2]) & padInd[2].ult(pad[2] + srcDims[2]);
+
+  Expr output = Expr::mkIte(cond, input.get(srcInd).first, *getZero(elemTy))
+                + bias.get({padInd[3]}).first;
+  Expr init = Expr::mkIte(cond, Expr::mkBool(true), input.isInitialized(padInd));
+
+  auto padInput = Tensor::mkLambda(
```
aqjune, line 1012: You can use `Tensor::mkInitializedLambda` here

aqjune: Also, please add `st.wellDefined(op, {input, weight, bias}.isFullyInitialized())`

tests/long-opts/conv2d-nchw/convolution_nchw.src.mlir
```diff
@@ -1,10 +0,0 @@
-// VERIFY
```
aqjune, line 1: Is there a reason why this test has been removed?

inwchun: oops. forgot to add file after renaming dir

PR #206, opened 2021-12-02
Improve fp bitwidth calculation policy

strikef:
This PR introduces a new FP bitwidth calculation method.
- If FP casting abstraction level is `FULLY_ABS`, each `AbsFpEncoding` is constructed as following:
  - Both limit and precision bitwidth are set to 0.
  - Smaller value bitwidth is calculated as `log2_ceil(#variables + #constants + 2)`
- If FP casting abstraction level is `PRECISE`, there's a whole new policy to decide each bitwidth. It is somewhat complex to explain in PR description, so please refer to paper repo for more information.

### DIFF
--- a/src/abstractops.cpp
+++ b/src/abstractops.cpp
@@ -102,9 +102,9 @@ void setAbstraction(
   // + 2: reserved for +NaN, +Inf; separately counted because they cannot be
   // included in set<APFloat>
   unsigned floatBits = log2_ceil(floatNonConstsCnt + floatConsts.size() + 2);
-  unsigned doubleBits = log2_ceil(doubleNonConstsCnt + doubleConsts.size() + 2);
+  floatEnc.emplace(llvm::APFloat::IEEEsingle(), floatBits, "float");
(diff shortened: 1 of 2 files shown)


src/abstractops.cpp
```diff
@@ -275,20 +269,20 @@ FnDecl AbsFpEncoding::getUltFn() {
...
-    fp_extendfn.emplace({fty}, fty, "fp_extend_" + fn_suffix);
+    auto src_fty = Sort::bvSort(fp_bitwidth);
+    auto tgt_fty = Sort::bvSort(tgt.fp_bitwidth);
+    fp_extendfn.emplace({src_fty}, tgt_fty, "fp_extend_" + fn_suffix);
   }
   return *fp_extendfn;
 }
 
-FnDecl AbsFpEncoding::getTruncateFn() {
+FnDecl AbsFpEncoding::getTruncateFn(const AbsFpEncoding &tgt) {
```
aqjune, line 281: Let's remove `AbsFpEncoding::smaller_fpty_enc`

PR #210, opened 2021-12-04
Use a unique hashing function when encoding sum_fn's equality

makesource:
Based on previous analysis, we can verify all cases by using a unique hashing function.


src/abstractops.cpp
```diff
@@ -777,7 +777,7 @@ Expr AbsFpEncoding::getFpAssociativePrecondition() const {
       if (!an.isUInt(alen) || !bn.isUInt(blen) || alen != blen) continue;
 
       auto domainSort = a.select(Index(0)).sort();
-      FnDecl hashfn(domainSort, Index::sort(), freshName("fp_hash"));
+      FnDecl hashfn(domainSort, Index::sort(), "fp_hash");
```
aqjune, line 780: Should we change the range (치역) of fp_hash?

makesource: The range is already Index::sort() (which is BV(32))
Should it be more optimized or determined dynamically?

aqjune:
I think so. Tensor size is guaranteed to be within Index::sort, but not m^2
n is.

2021년 12월 4일 (토) 오후 11:51, seongwon bang ***@***.***>님이 작성:

src/abstractops.cpp
aqjune: Another comment: You'll want to declare FnDecl only once & reuse it. (e.g. AbsFpEncoding::getAddFn)

PR #212, opened 2021-12-05
Implement FP division

strikef:
This PR implements abstract FP division (aka `smt::Expr AbsFpEncoding::div()`)
- If one or more operand of div is NaN, it yields NaN as well.
- 0.0 / 0.0 and Inf / Inf yields NaN.
- finite x / Inf yields 0.0
- Inf / finite x yields Inf
- Finite x / 0.0 yields Inf.  
This is division by zero and signaling exception should be raised according to IEEE754 standard, but LLVM ignores the signal and continues the execution. We follow this semantics as well.
- Otherwise, it yields an arbitrary BV with appropriate sign.
- Litmus tests are added to verify the above behaviors


aqjune:
Q: what is the relation between div and mul?

PR #213, opened 2021-12-06
Add basic sum precondition

makesource:
Add precondition for `sum([a]) = a`


makesource:
I'll update more simpler version

PR #217, opened 2021-12-07
Implement precise FP rounding

strikef:
This PR implements precondition-based rounding FP truncation.

- The values being truncated are now properly 'rounded' instead of being 'floored`.
- FP truncation is now implemented based on UF and preconditions
  - FP constant analysis adds preconditions, which limits the possible value of each truncated constant to a single value
  - `AbsFpEncoding::truncate()` adds preconditions, which limits the possible value of each truncated variable to 'floored' and 'ceiled' value. 
  - In case there are more than one possible truncation result, other preconditions may be added in the future to choose one of them.
- Fixed several bugs in previous truncation behaviors
- Litmus tests are added to verify the behaviors described above


aqjune:
I'll merge this PR, but could you make another PR that contains more tests?

PR #225, opened 2021-12-11
[WIP] Dynamic FP constant mapping

strikef:
This PR implements dynamic FP constant mapping, using the named variables and preconditions

## Static constant mapping
FP constant mapping used to be static. Every constant that appears in IR, or constant that should exist in the encoding are first ordered. BVs are then assigned to each constant in an incremental manner. This means we can in theory calculate the BV value of every constant that appears in any given IR. The behavior is quite straightforward and can be easily implemented.

## Limitations of static mapping
However, the possibility of 'the values between' is completely ignored in this algorithm. Most float constants do have one or more values between them, but the incremental static mapping virtually eliminates this gap and leaves no space for these values.

## Solution: precondition-based mapping
In this PR, each constant is assigned a set of BV variables. Precise casting is supported with help of named variables, and fresh variables allow more arbitrary value assignment. Then, a total order is given between abstract representation of constants using preconditions, like `-Inf < min_const < ... < -0.0 < 0.0 < ... <max_const < Inf`. If precise truncation becomes necessary, another set of preconditions are added to fix the direction of rounding for each constant when a truncation occurs.

## Known issues (TOFIX)
- This feature breaks counterexample printing. The actual value inside a variable is unknown unless it is evaluated against the model, so the equality check between abstract representation and `value::Float` fails.
- Validating test cases that use a lot of non-constant FPs takes much longer. As of now, validating `litmus/tensor-ops/extract_tensor_sum` times out  due to this issue.


aqjune:
https://github.com/aqjune/mlir-tv/commit/5b3977e81315ec0364f4106d708d3f57612938a7 fixes the regression in dot_rewrite_manually_assoc.src.mlir!

Now, please fix the C'Ex test and I will merge this PR.

PR #226, opened 2021-12-13
Refactor abstraction refinement

makesource:


aqjune:
vinput.dumpSMTPath update is missing - I will fix that.

src/vcgen.cpp
```diff
@@ -499,120 +500,94 @@ static Results validate(ValidationInput vinput) {
...
-      << "  AbsLevelFpDot: " << magic_enum::enum_name(afd) << "\n"
-      << "  AbsLevelFpCast: " << magic_enum::enum_name(afc) << "\n"
-      << "  AbsLevelIntDot: " << magic_enum::enum_name(aid) << "\n"
-      << "  AbsFpAddSumEncoding: " << magic_enum::enum_name(fas) << "\n"
+      << "  Current semantics of abstractly defined ops..\n"
+      << "  AbsLevelFpDot: " << magic_enum::enum_name(level.alFpDot) << "\n"
+      << "  AbsLevelFpCast: " << magic_enum::enum_name(level.alFpCast) << "\n"
+      << "  AbsLevelIntDot: " << magic_enum::enum_name(level.alIntDot) << "\n"
+      << "  AbsFpAddSumEncoding: " << magic_enum::enum_name(level.fpAddSumEncoding) << "\n"
+      << "  Verification Result: " << magic_enum::enum_name(result.code) << "\n"
```
aqjune, line 511: Will make this line print only when -verbose is given.

aqjune:
I think using priority_queue isn't needed at the moment. Will switch to simply using queue.

PR #229, opened 2021-12-17
Implement to support tosa.depthwise_conv2d operation

inwchun:
implement for depthwise op


src/encode.cpp
```diff
@@ -1035,7 +1136,7 @@ void encodeOp(State &st, mlir::tosa::Conv2DOp op, bool) {
 
   auto output = Tensor::mkInitializedLambda(
                   elemTy, move(outDims), move(outInd), 
-                  biasf.add(tf)
+                  tf.add(biasf)
```
aqjune, line 1139: Why is this changed?

aqjune:
You'll need to update `analyzeBlock` in `analysis.cpp` as well.

aqjune:
Can we merge common parts of this and `encodeOp(State &st, mlir::tosa::Conv2DOp op, ..)`?

src/encode.cpp
inwchun: to be same as the tosa spec

aqjune:
Build is broken?

aqjune:
Oops, the failures are from Z3 API updates. I will update them.

aqjune:
Ehh, no, they are compile failures in the Z3 header.

src/encode.cpp
aqjune: Okay, fine.
FYI: since fp add is commutative, they are equivalent.

inwchun:
I think it's an internal error in z3

PR #230, opened 2021-12-17
Implement cmpf operator

strikef:
This PR implements comparison operator for float types.

- cmp is implemented as a set of concrete ite expressions. Each expression yields a BV(1) from two BV(fp_bitwidth)s.
- Using cmp looks like this: `cmp(pred, f1, f2)`
- `pred` is an enum class `aop::CmpPredicate`. There are 16 members in `aop::CmpPredicate`, and each represents the corresponding cmpf predicate in MLIR. (`[Ordered|Unordered][EQ|NE|LE|LT|GE|GT], Ordered, Unordered, True, False`) 
- Litmus tests are added to verify the behavior.

### DIFF
--- a/src/abstractops.cpp
+++ b/src/abstractops.cpp
@@ -943,7 +943,7 @@ Expr AbsFpEncoding::dot(const Expr &a, const Expr &b, const Expr &n) {
   llvm_unreachable("Unknown abstraction level for fp dot");
 }
 
-Expr AbsFpEncoding::cmp(const CmpPredicate pred,
+Expr AbsFpEncoding::cmp(mlir::arith::CmpFPredicate pred,
                         const Expr &f1, const Expr &f2) {
   const Expr trueBV = Expr::mkBV(1, 1);
   const Expr falseBV = Expr::mkBV(0, 1);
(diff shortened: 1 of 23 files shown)


src/abstractops.cpp
```diff
@@ -959,8 +949,64 @@ Expr AbsFpEncoding::dot(const Expr &a, const Expr &b, const Expr &n) {
...
+                        trueBV, falseBV))));
+  const Expr cmpGT = Expr::mkIte(f1Sign.ult(f2Sign), trueBV,
+                      Expr::mkIte(f1Sign.ugt(f2Sign), falseBV,
+                      Expr::mkIte(f1Magn == f2Magn, falseBV,
+                      Expr::mkIte(f1Magn.ugt(f2Magn) ^ f1Sign.isNonZero(),
+                        trueBV, falseBV))));
+
+  switch (pred) {
+  case CmpPredicate::OEQ:
+    return Expr::mkIte(hasNaN, falseBV, cmpEQ);
```
aqjune, line 976: Let's use `!hasNaN & cmpEQ`

src/abstractops.cpp
```diff
@@ -959,8 +949,64 @@ Expr AbsFpEncoding::dot(const Expr &a, const Expr &b, const Expr &n) {
...
+  case CmpPredicate::OLE:
+    return Expr::mkIte(hasNaN, falseBV, cmpEQ | cmpLT);
+  case CmpPredicate::OLT:
+    return Expr::mkIte(hasNaN, falseBV, cmpLT);
+  case CmpPredicate::OGE:
+    return Expr::mkIte(hasNaN, falseBV, cmpEQ | cmpGT);
+  case CmpPredicate::OGT:
+    return Expr::mkIte(hasNaN, falseBV, cmpGT);
+  case CmpPredicate::UEQ:
+    return Expr::mkIte(hasNaN, trueBV, cmpEQ);
```
aqjune, line 988: Let's use `(!hasNaN).implies(cmpEQ)`

src/abstractops.h
```diff
@@ -205,9 +212,9 @@ class AbsFpEncoding {
   smt::Expr sum(const smt::Expr &a, const smt::Expr &n);
   smt::Expr exp(const smt::Expr &x);
   smt::Expr dot(const smt::Expr &a, const smt::Expr &b, const smt::Expr &n);
-  smt::Expr fult(const smt::Expr &f1, const smt::Expr &f2);
   smt::Expr extend(const smt::Expr &f, aop::AbsFpEncoding &tgt);
   smt::Expr truncate(const smt::Expr &f, aop::AbsFpEncoding &tgt);
+  smt::Expr cmp(const CmpPredicate pred, const smt::Expr &f1, const smt::Expr &f2);
```
aqjune, line 217: What do you think about directly accepting `mlir::arith::CmpFPredicate`?
```
  smt::Expr cmp(const mlir::arith::CmpFPredicate pred, const smt::Expr &f1, const smt::Expr &f2);
```

PR #233, opened 2021-12-19
Add UB condition for tosa.conv2d and tosa.depthwise_conv2d

inwchun:
add ub op

### DIFF
--- a/src/encode.cpp
+++ b/src/encode.cpp
@@ -1008,12 +1008,15 @@ static Tensor getPaddedTensor2D(mlir::Type elemTy,
 static Tensor addBias2D(mlir::Type elemTy, 
                         vector<Expr> dims,
                         Tensor acc, Tensor bias) {
-  vector<Expr> outInd = Index::boundIndexVars(acc.getRank());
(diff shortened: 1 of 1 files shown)


src/encode.cpp
```diff
@@ -1063,14 +1068,20 @@ void encodeOp(State &st, mlir::tosa::DepthwiseConv2DOp op, bool) {
...
+
   // NxOHxOWx(C*M)
-  vector<Expr> outDims = {N, tDims[1], tDims[2], C * M};
+  vector<Expr> tDims = {N, t2DDims[1], t2DDims[2], C * M};
 
-  auto output = addBias2D(elemTy, outDims, t, bias);
+  auto acc = Tensor::mkInitializedLambda(
+            elemTy, move(tDims), move(outInd), 
+            t2D.get({Index(0), outInd[1], outInd[2], Index(0)}).first
+          );
```
aqjune, line 1082: Why is this added?

inwchun: there is an error in the previous version... forgot to push the commit. The index should be different since t2D is 1xOHxOWx1

PR #238, opened 2021-12-20
Implement linalg.depthwise_conv_2d_nhwc_hwcm

inwchun:
implement linalg.depthwise_conv_2d_nhwc_hwcm

### DIFF
--- a/src/encode.cpp
+++ b/src/encode.cpp
@@ -1034,53 +1034,14 @@ void encodeOp(State &st, mlir::tosa::DepthwiseConv2DOp op, bool) {
 
   auto elemTy = getElemTy(op.getResult());
 
+  // Check whether C is identical
+  st.wellDefined(op, input.getDim(3) == weight.getDim(2));
+  // Check whether C * M is identical
+  st.wellDefined(op, bias.getDim(0) == (weight.getDim(2) * weight.getDim(3)));
+
   auto paddedTensor = getPaddedTensor2D(elemTy, input, op.pad());
 
-  vector<Expr> outInd = Index::boundIndexVars(4);
-  auto wDims = weight.getDims();
-  auto padDims = paddedTensor.getDims();
-  auto N = padDims[0];
-  auto C = wDims[2];
-  auto M = wDims[3];
-
-  // Check whether C is identical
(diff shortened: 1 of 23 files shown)


aqjune:
I'll look into this tomorrow.

src/encode.cpp
```diff
@@ -685,6 +685,7 @@ void encodeOp(State &st, mlir::AffineApplyOp op, bool) {
 
 template<>
 void encodeOp(State &st, mlir::ReturnOp op, bool) {
+  llvm::outs() << op.getOperand(0) << "\n";
```
aqjune, line 688: Remove dbg?

src/encode.cpp
```diff
@@ -1034,52 +1035,24 @@ void encodeOp(State &st, mlir::tosa::DepthwiseConv2DOp op, bool) {
```
aqjune, line 1035: Could you make a separate PR that only updates `encodeOp(State &st, mlir::tosa::DepthwiseConv2DOp ...)` please?
I will merge the PR first once it is ready.

aqjune: Let's make a PR that update encoding `mlir::tosa::DepthwiseConv2DOp` to use `getPaddedTensor2D` first.

src/value.cpp
```diff
@@ -546,6 +546,53 @@ Tensor Tensor::concat(const Tensor &t2, size_t axis) {
       move(dim), move(idx), move(elem));
 }
 
+Tensor Tensor::depthwiseConv2D(const Tensor &filter,
+    const vector<Expr> &strides,
+    const vector<Expr> &dilations) const {
+
+  assert(getDims().size() == 4);
```
aqjune, line 553: Could you leave assertions on `filter` and `dilations`?

src/value.h
```diff
@@ -177,6 +177,12 @@ class Tensor: public ShapedValue {
       const std::vector<smt::Expr> &dilations,
       ConvLayout layout) const;
 
+  // tensor which is depthwise convolution of this 2D tensor and filter.
```
aqjune, line 180: Please leave a comment saying that callers of this fn must check whether the tensors are initialized (look `conv()` above).

PR #244, opened 2021-12-21
Fix reduction loop to consider output's initial value

makesource:
- `linalg::generic` reduction loop consider output tensor's initial value
- `linalg::dot`, `linalg::matmul` also consider output tensor's initial value
- `linalg::conv`, `linalg::DepthwiseConv2DNhwcHwcmOp` also consider output tensor's initial value


src/encode.cpp
```diff
@@ -2290,11 +2294,13 @@ void encodeOp(State &st, mlir::linalg::DotOp op, bool encodeMemWrite) {
...
   auto t1 = st.regs.get<Tensor>(inputOps[0]->get());
   auto t2 = st.regs.get<Tensor>(inputOps[1]->get());
+  auto t3 = st.regs.get<Tensor>(outputOps[0]->get());
   st.wellDefined(op, t1.isFullyInitialized());
   st.wellDefined(op, t2.isFullyInitialized());
+  st.wellDefined(op, t3.isFullyInitialized());
   st.wellDefined(op, t1.get1DSize() == t2.get1DSize());
 
-  auto res = t1.dot(t2);
+  auto res = t1.dot(t2, move(t3));
```
aqjune, line 2303: Let's use `t3.get(Index(0))` for clarity?

src/encode.cpp
```diff
@@ -2613,8 +2617,13 @@ static void encodeReductionLoopBodyAndOutput(
     // t_res[0] = sum(\i. t_input[i / n][i % n] , i < m * n)
 
     // Define this as a splat tensor (num. elems is 1 anyway)
-    t_res = Tensor(t_v.getElemType(), t_v.sum(),
+    // TODO (seongwon): Currently we cover only tensor cases..
+    // TODO: Support memref cases
+    auto outty = newst.regs.get<Tensor>(the_op->getOperands().back());
```
aqjune, line 2622: Nit: `outty` vs. `outTensor`?

aqjune: Another question: can a reduction loop have multiple outputs?

src/encode.cpp
```diff
@@ -2613,8 +2617,13 @@ static void encodeReductionLoopBodyAndOutput(
     // t_res[0] = sum(\i. t_input[i / n][i % n] , i < m * n)
 
     // Define this as a splat tensor (num. elems is 1 anyway)
-    t_res = Tensor(t_v.getElemType(), t_v.sum(),
+    // TODO (seongwon): Currently we cover only tensor cases..
+    // TODO: Support memref cases
```
aqjune, line 2621: I thought linalg.generic for memref was already supported - no?

src/value.cpp
```diff
@@ -685,13 +685,21 @@ Tensor Tensor::matmul(const Tensor &b) const {
   auto bt_row = bt.to1DArrayWithOfs(
       {j, Index::zero()}, {Index::one(), bt.dims[1]});
 
-  auto res = elemType.isa<mlir::FloatType>() ?
+  optional<Expr> res;
+  if (init) {
+    auto initVal = init->get({i, j}).first;
+    res = elemType.isa<mlir::FloatType>() ?
```
aqjune, line 691: Maybe we can merge two bodies below?

src/value.cpp
```diff
@@ -720,17 +728,26 @@ Tensor Tensor::elementwiseUnaryOp(
   return mkInitializedLambda(resultElemType, getDims(), move(idxvars), elemout);
 }
 
-Expr Tensor::dot(const Tensor &t2) const {
+Expr Tensor::dot(const Tensor &t2, std::optional<Tensor> &&init) const {
```
aqjune, line 731: No std::

src/value.cpp
```diff
@@ -720,17 +728,26 @@ Tensor Tensor::elementwiseUnaryOp(
...
   auto len = get1DSize();
-  return elemType.isa<mlir::FloatType>() ?
+
+  if (init) {
+    // scalar output
+    auto initVal = init->get({Index(0)}).first;
+    return elemType.isa<mlir::FloatType>() ?
```
aqjune, line 737: Let's merge

aqjune:
Do you want to add a test that is verify-incorrect due to different out args?

src/encode.cpp
makesource: It checked at here (https://github.com/aqjune/mlir-tv/blob/master/src/encode.cpp#L2704-L2706)

src/encode.cpp
makesource: Partially yes but.. to properly support reduction loop, we have to check whether output memref is initialized.

makesource:
Added `matmul-bad` cases. Thanks!
https://github.com/aqjune/mlir-tv/pull/244/files#diff-7cd8f871ada445d1529adf11924aa01a34b859a4f9b15e5bdbb8058fcbf4eadd

PR #249, opened 2021-12-22
Implement missing semantics for fp div

strikef:
This PR implements two important properties of FP division (which was missing)
- +-0.0 / x -> ?0.0 (x != +-0.0)
- x / x -> 1.0, x / -x -> -1.0 (x != +-0.0 or +-Inf)


aqjune:
Q: what about nan inputs?

strikef:
If dividend or divisor is NaN, the result is NaN. This condition is checked before checking the above two conditions.

PR #251, opened 2021-12-22
Optimize convolution operations

inwchun:
optimize convolution operation

### DIFF
--- a/src/smt.cpp
+++ b/src/smt.cpp
@@ -119,8 +119,12 @@ vector<Expr> from1DIdx(
   // Start from the lowest dimension
   for (size_t ii = dims.size(); ii > 0; --ii) {
     size_t i = ii - 1;
-    idxs.emplace_back(i == 0 ? idx1d : idx1d.urem(dims[i]));
-    idx1d = idx1d.udiv(dims[i]);
+    uint64_t v;
+    auto isOne = dims[i].isUInt(v) && v == 1;
+
+    // how to make Expr(0)?
+    idxs.emplace_back(i == 0 ? idx1d : (isOne ? Expr::mkBV(0, 32U) : idx1d.urem(dims[i])));
+    idx1d = isOne ? idx1d : idx1d.udiv(dims[i]);
   }
 
   reverse(idxs.begin(), idxs.end());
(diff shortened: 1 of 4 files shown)


src/smt.cpp
```diff
@@ -119,8 +119,12 @@ vector<Expr> from1DIdx(
...
   for (size_t ii = dims.size(); ii > 0; --ii) {
     size_t i = ii - 1;
-    idxs.emplace_back(i == 0 ? idx1d : idx1d.urem(dims[i]));
-    idx1d = idx1d.udiv(dims[i]);
+    uint64_t v;
+    auto isOne = dims[i].isUInt(v) && v == 1;
+
+    // how to make Expr(0)?
+    idxs.emplace_back(i == 0 ? idx1d : (isOne ? Expr::mkBV(0, 32U) : idx1d.urem(dims[i])));
+    idx1d = isOne ? idx1d : idx1d.udiv(dims[i]);
```
aqjune, line 127: This is not necessary because Expr::udiv already does the optimization

src/smt.cpp
```diff
@@ -119,8 +119,12 @@ vector<Expr> from1DIdx(
   // Start from the lowest dimension
   for (size_t ii = dims.size(); ii > 0; --ii) {
     size_t i = ii - 1;
-    idxs.emplace_back(i == 0 ? idx1d : idx1d.urem(dims[i]));
-    idx1d = idx1d.udiv(dims[i]);
+    uint64_t v;
+    auto isOne = dims[i].isUInt(v) && v == 1;
+
+    // how to make Expr(0)?
+    idxs.emplace_back(i == 0 ? idx1d : (isOne ? Expr::mkBV(0, 32U) : idx1d.urem(dims[i])));
```
aqjune, line 126: You can use `Expr::mkBV(0, dims[0])` to use `dims[0]`'s sort.
BTW, does this improve performance?

src/smt.cpp
inwchun: I think the Expr::urem should be optimized? x%1==0

src/smt.cpp
inwchun: printed expr and found (urem var #0001). isn't this 0?

PR #253, opened 2021-12-27
Fix fp identity in to1DArrayWithOfs

makesource:
`zero()` returns non-negative zero value


aqjune:
Thanks! I'll merge this once CI is passed.

PR #256, opened 2021-12-27
Optimize lambda expression

makesource:


makesource:
I'm not sure whether this patch improves performance, I'll create a new PR.

PR #268, opened 2022-01-03
Fix issue where truncate() may unexpectedly yield NaN

strikef:
Due to a pitfall in our FP encoding, truncating a value of which TB == INT_MAX yields NaN although the original value is some finite value.
Therefore, even if LB == 0, truncate() now explicitly yields Inf if the TB >= INT_MAX - 1.
Also, constants beyond LFP's range may be assigned a BV of which LB == 0 and TB >= INT_MAX - 1.

### DIFF
--- a/src/abstractops.cpp
+++ b/src/abstractops.cpp
@@ -1100,7 +1100,8 @@ Expr AbsFpEncoding::truncate(const smt::Expr &f, aop::AbsFpEncoding &tgt) {
   return Expr::mkIte(isnan(f), tgt.nan(),
           Expr::mkIte(f == infinity(), tgt.infinity(),
           Expr::mkIte(f == infinity(true), tgt.infinity(true),
-          Expr::mkIte(limit_bits != limit_zero,
+          Expr::mkIte(limit_bits != limit_zero |
+                      floored_value.uge(tgt.getMagnitudeBits(tgt.infinity())),
             Expr::mkIte(sign_bit == sign_pos,
               tgt.infinity(), tgt.infinity(true)),
             Expr::mkIte(is_prec_zero, floored_float,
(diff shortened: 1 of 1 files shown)


src/abstractops.cpp
```diff
@@ -438,8 +438,12 @@ void AbsFpEncoding::addConstants(const set<llvm::APFloat>& const_set) {
         auto sv_prec_bits = Expr::mkFreshVar(Sort::bvSort(sv_prec_bitwidth),
                             "fp_const_sval_prec_bits_");
 
-        limit_bits = mkNonzero(
-            Expr::mkFreshVar(limit_bits, "fp_const_limit_bits_"));
+        auto inf_sv_value = (1 << sv_prec_bitwidth) -
+                            (1 << value_bit_info.prec_bitwidth);
+        auto limit_var = Expr::mkFreshVar(limit_bits, "fp_const_limit_bits_");
+        limit_bits = Expr::mkIte(sv_prec_bits.uge(inf_sv_value), 
+          limit_var,
```
aqjune, line 445: Why is this var not nonZeroed?

src/abstractops.cpp
```diff
@@ -1100,7 +1104,8 @@ Expr AbsFpEncoding::truncate(const smt::Expr &f, aop::AbsFpEncoding &tgt) {
   return Expr::mkIte(isnan(f), tgt.nan(),
           Expr::mkIte(f == infinity(), tgt.infinity(),
           Expr::mkIte(f == infinity(true), tgt.infinity(true),
-          Expr::mkIte(limit_bits != limit_zero,
+          Expr::mkIte(limit_bits != limit_zero |
+                      floored_value.uge(tgt.getMagnitudeBits(tgt.infinity())),
```
aqjune, line 1108: Could you factor out `floored_value.uge(tgt.getMagnitudeBits(tgt.infinity())` as a separate variable please?

src/abstractops.cpp
strikef: Consider this case: let's assume that f32_max + epsilon has been assigned the BV 0[SB] 01[LB] 00000[TB] 00[PB]. Then, BVs within and including the range (0 00 11110 00 .. 0 00 11111 11) may represent invalid values that does not exist in real-world double. Therefore we should assign the constants starting from such BVs.

PR #270, opened 2022-01-04
Support 4-dimension transpose

makesource:


src/value.cpp
```diff
@@ -545,18 +545,11 @@ Tensor Tensor::affine(
...
   return {
     elemType,
     move(newsizes),
-    Expr::mkLambda( // Value
-      idxvar,
-      Expr::mkIte(
-        ((Expr)idxvar).ult(::get1DSize(newsizes)), // TODO: is this chk needed?
-        elem,
-        identity
-      )),
```
aqjune, line 559: I will make a separate commit for this part to see the performance impact in CI

PR #271, opened 2022-01-05
Support matmul with memref operands

makesource:


makesource:
I moved `fold-tensor-extract-op/var` test cases to long-opts because it takes about 60 secs when AUFBV logic given.
(It actually has forall quantifier in smt expression but... it was solved more quickly when QF_AUFBV logic given. I think it's an unintended behavior, I fixed it to run using AUFBV logic and moved to long-opts)

src/encode.cpp
```diff
@@ -1597,28 +1599,46 @@ void encodeOp(State &st, mlir::tensor::ExpandShapeOp op, bool) {
...
-        "tensor semantics is supported only");
+        "tensor/buffer semantics is supported only");
 
   if (op.getNumInputs() != 2 || op.getNumOutputs() != 1)
     throw UnsupportedException(op.getOperation(),
         "unsupported form");
 
   if (getElemTy(op.getOperand(0)) != getElemTy(op.getOperand(1)) ||
-      getElemTy(op.getOperand(0)) != getElemTy(op.getResult(0)))
+      getElemTy(op.getOperand(0)) != getElemTy(op.getOperand(2)))
```
aqjune, line 1611: Good catch!

src/encode.cpp
```diff
@@ -1597,28 +1599,46 @@ void encodeOp(State &st, mlir::tensor::ExpandShapeOp op, bool) {
...
+    auto mc = st.regs.get<MemRef>(op.getOperand(2));
+    auto aTy = op.getOperand(0).getType().cast<mlir::MemRefType>();
+    auto bTy = op.getOperand(1).getType().cast<mlir::MemRefType>();
+    auto cTy = op.getOperand(2).getType().cast<mlir::MemRefType>();
+    Tensor a = loadTensor(st, op, ma, aTy);
+    Tensor b = loadTensor(st, op, mb, bTy);
+    Tensor c = loadTensor(st, op, mc, cTy);
+    Tensor result = a.matmul(b, /*transposed*/false, c);
+
+    storeTensorTo(st, op, move(result), mc, cTy, true);
```
aqjune, line 1640: Let's add the aliasing check: a noalias c /\ b noalias c.

src/encode.cpp
```diff
@@ -1597,28 +1599,46 @@ void encodeOp(State &st, mlir::tensor::ExpandShapeOp op, bool) {
 
 template<>
 void encodeOp(State &st, mlir::linalg::MatmulOp op, bool) {
```
aqjune, line 1601: You'll need to check the third argument (`encodeMemWriteOp`).

src/encode.cpp
makesource: Added no alias check `mc.noalias(ma) & mc.noalias(mb)`

PR #272, opened 2022-01-05
Support memref.collapse_shape, memref.expand_shape operation

makesource:


src/value.cpp
```diff
@@ -1608,6 +1608,11 @@ MemRef MemRef::subview(const vector<Expr> &offsets,
   }
 }
 
+MemRef MemRef::reshape(const std::vector<smt::Expr> &newDims) {
```
aqjune, line 1611: Should we insert an assertion stating that this MemRef has an identity layout?

src/value.cpp
```diff
@@ -1608,6 +1608,11 @@ MemRef MemRef::subview(const vector<Expr> &offsets,
...
+  return MemRef(m, elemType, bid, offset,
+    newDims, MemRef::Layout(newDims), Expr::mkBool(true));
```
aqjune, line 1613: Let's leave a comment in its declaration (value.h) saying that the returned MemRef is a view reference.

src/value.cpp
makesource: resolved

src/value.cpp
makesource: resolved

PR #275, opened 2022-01-06
Fix crashes when run CVC5 only

makesource:


src/smt.cpp
```diff
@@ -617,21 +620,37 @@ Expr Expr::select(const vector<Expr> &idxs) const {
...
+    }
   }
-  if (Lambda(Any(elem)).match(arr) && idxs.size() == 1) {
-    e = elem->substituteDeBruijn({idxs[0]});
+#endif // SOLVER_Z3
+#ifdef SOLVER_CVC5
+  if (e.hasCVC5Term()) {
+    Expr arr;
+    arr.setCVC5(e.getCVC5Term()[0]); // get body term
+
```
aqjune, line 643: Can we merge the part below with line 629~637?
Matchers are solver-agnostic.

src/smt.cpp
```diff
@@ -406,9 +406,12 @@ bool Expr::isTrue() const {
...
 bool Expr::isVar() const {
   bool res = false;
-  IF_Z3_ENABLED(res |= z3 && z3->is_app() && z3->is_const() &&
-                       !z3->is_numeral());
-  // TODO: CVC5
+  IF_Z3_ENABLED(
+    if (z3) res |= z3->is_app() && z3->is_const() && !z3->is_numeral()
+  );
+  IF_CVC5_ENABLED(
+    if (cvc5) res |= !(cvc5->isIntegerValue() || cvc5->isBitVectorValue())
```
aqjune, line 413: We also have boolean constants.
BTW, is there a better way than this? We may have FPA constants at some point as well.

src/smt.cpp
```diff
@@ -1533,155 +1551,262 @@ void Matcher::setZ3(Expr &e, optional<z3::expr> &&opt) const {
...
+    auto e = expr.getCVC5Term();
+
+    if (!e.isConstArray())
+      return false;
 
-  Expr newe = newExpr();
-  IF_Z3_ENABLED(setZ3(newe, z3::expr(*sctx.z3, Z3_get_app_arg(*sctx.z3, a, 0))));
-  return subMatcher(newe);
+    Expr newe = newExpr();
+    setCVC5(newe, e.getConstArrayBase());
```
aqjune, line 1628: I think you should call `setCVC5` from `newe` updated at line 1616.

aqjune:
Nice! There definitely was significant progress from the previous version.
Let's update the PRs so that matchers' argument Expr contain both Z3 expr & CVC5 expr (as I commented above)

src/smt.cpp
makesource: Well... I think it's impossible to match Z3/CVC5 terms at the same time because the expression can be different.
So I think we should separate the implementation of Z3/CVC5 matchers on the caller's side.

src/smt.cpp
makesource: I'll update to consider FPA and boolean constants as well.
I couldn't find a proper api to test whether the term is variable :(

src/smt.cpp
aqjune: Okay.. I think the complexity comes from the fact that mlir-tv can run with two solvers (Z3, CVC5) enabled at runtime.
However, we are currently not getting any benefit from it; even if one of them finishes early, we cannot stop the other.
I will remove https://github.com/aqjune/mlir-tv/blob/17d7a7ebad2d2e000a49d5ee55dca3ee1e231dac/src/main.cpp#L53 so that only one of them is enabled at runtime. What do you think?

makesource: Look good to me!! I'll update it.

PR #283, opened 2022-01-11
Implement bag union and add related precondition

makesource:


src/smt.cpp
```diff
@@ -744,6 +744,24 @@ Expr Expr::insert(const Expr &elem) const {
...
+Expr Expr::bagUnion(const Expr &other) const {
+  CHECK_LOCK2(other);
+    Expr e;
+  // Z3 doesn't support multisets. We encode it using a const array.
+  SET_Z3(e, fupdate2(sctx.z3, z3, [&](auto &ctx, auto &arrayz3) {
+    auto domain = arrayz3.get_sort().array_domain();
+    auto idx = ctx.constant("idx", domain);
+    auto lhs = z3::select(arrayz3, idx);
+    auto rhs = z3::select(*other.z3, idx);
+    return z3::lambda(idx, lhs + rhs);
```
makesource, line 756: z3 internally treat lambda and array as the same type. 
So both types are compatible.

PR #284, opened 2022-01-12
Analyze OP intermediate fp values

makesource:


aqjune:
Could you fix the CI failure please?

PR #298, opened 2022-01-17
Fix crashes due to use of wrong cvc5 Term

strikef:
This PR fixes crashes due to lack of sort check procedures in `Expr::mkVar()`.
Previously it only checked the variable name when fetching the cached `cvc5::Term`, but this caused problems when two `cvc5::Term`s of different sort shared the same name. From now on, `Expr::mkVar()` checks both sort and name when fetching the cached `cvc5::Term`.


strikef:
Disregard: planning on implementing different semantics

PR #305, opened 2022-01-20
Add zero-dimension tensor support for `mlir::tensor::FromElementsOp`

strikef:
This PR fixes the issue where encoding `mlir::tensor::FromElementsOp` crashed upon encountering zero-dim tensor situation.
If the operand is not given and the return type is 0-dim tensor, 0-dim `Tensor` of Op return type is created

### DIFF
--- a/src/encode.cpp
+++ b/src/encode.cpp
@@ -1938,14 +1938,8 @@ void encodeOp(State &st, mlir::tensor::FromElementsOp op, bool) {
   auto elemTy = op.getType().getElementType();
   if (elems.size() == 0) {
     // Zero-dim tensor: derive Sort from Op return type
-    optional<Sort> elemSort;
-    if (elemTy.isa<mlir::FloatType>())
-      elemSort = Float::sort(elemTy);
-    else if (elemTy.isIndex())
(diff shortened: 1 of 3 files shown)


src/encode.cpp
```diff
@@ -1936,7 +1936,21 @@ void encodeOp(State &st, mlir::tensor::FromElementsOp op, bool) {
...
+    if (elemTy.isa<mlir::FloatType>())
+      elemSort = Float::sort(elemTy);
+    else if (elemTy.isIndex())
+      elemSort = Index::sort();
+    else if (elemTy.isInteger(elemTy.getIntOrFloatBitWidth()))
+      elemSort = Integer::sort(elemTy.getIntOrFloatBitWidth());
+    else
+      throw UnsupportedException(op.getOperation(), "Unsupported type");
+    st.regs.add(op.getResult(), Tensor(elemTy, move(*elemSort)));
+  }
```
makesource, line 1951: How about using `convertPrimitiveTypeToSort` ?

strikef: My gosh this is superduper elegant

PR #306, opened 2022-01-20
Regard memref with no dimension as scalar

strikef:
This PR fixes the issue with https://github.com/aqjune/mlir-tv/pull/288.
A memref with no explicit dimensions should be regarded as a scalar memref, not a zero-dim memref.


strikef:
The implementation was wrong... closing

Issue #312, opened 2022-02-02
Compilation failed with the newest LLVM build (PadTensorOP Removed)


Hatsunespica:
Looks like they removed this Op in the newest LLVM build. 
PadTensorOp was also removed from the doc
https://mlir.llvm.org/doxygen/structmlir_1_1linalg_1_1PadTensorOpTransformationPattern.html

below is the error message.

/home/spica/GitRepo/mlir-tv/src/analysis.cpp:295:30: error: ‘PadTensorOp’ is not a member of ‘mlir::linalg’; did you mean ‘InitTensorOp’?
  295 | bool analyzeOp(mlir::linalg::PadTensorOp op, AnalysisResult &res) {
      |                              ^~~~~~~~~~~
      |                              InitTensorOp
/home/spica/GitRepo/mlir-tv/src/analysis.cpp:295:61: error: expected primary-expression before ‘&’ token
  295 | bool analyzeOp(mlir::linalg::PadTensorOp op, AnalysisResult &res) {
      |                                                             ^
/home/spica/GitRepo/mlir-tv/src/analysis.cpp:295:62: error: ‘res’ was not declared in this scope
  295 | bool analyzeOp(mlir::linalg::PadTensorOp op, AnalysisResult &res) {
      |                                                              ^~~
/home/spica/GitRepo/mlir-tv/src/analysis.cpp:295:65: error: expression list treated as compound expression in initializer [-fpermissive]
  295 | bool analyzeOp(mlir::linalg::PadTensorOp op, AnalysisResult &res) {
      |                                                                 ^
/home/spica/GitRepo/mlir-tv/src/analysis.cpp:295:6: warning: too many template headers for ‘{anonymous}::analyzeOp’ (should be 0)
  295 | bool analyzeOp(mlir::linalg::PadTensorOp op, AnalysisResult &res) {
      |      ^~~~~~~~~
/home/spica/GitRepo/mlir-tv/src/analysis.cpp:295:66: error: expected ‘;’ before ‘{’ token
  295 | bool analyzeOp(mlir::linalg::PadTensorOp op, AnalysisResult &res) {
      |                                                                  ^~
      |

strikef:
Thank you for reporting this issue!
[...]

PR #313, opened 2022-02-03
Catch up with latest MLIR API

strikef:
This PR updates `mlir-tv` code and tests to catch up with latest MLIR API.
- `linalg.copy` is deprecated: `memref.copy` should be used instead.
- `TensorType::kDynamicSize` is removed: `ShapedType::kDynamicSize` should be used instead.
- `std.select` is moved to `arith.select` (https://github.com/llvm/llvm-project/commit/dec8af701ff25ea967d47d196e5247306754f8e8)
- `linalg.pad_tensor` is moved to `tensor.pad` (https://github.com/llvm/llvm-project/commit/fd0c6f53913f272ddd88948644fae36e63db120c)
- `OwningModuleRef` is replaced with `OwningOpRef<ModuleOp>` (https://github.com/llvm/llvm-project/commit/8f66ab1c2e00ca3e94dbdd2e435fbf0fe28bbee9)

Also, the automatic CI update script is re-enabled.


strikef:
#312

PR #316, opened 2022-02-09
fix link error during build

nbushehri:
Seems like building the tool with llvm trunk requires adding MLIRControlFlow library


aqjune:
It works great, thank you! :)

PR #319, opened 2022-02-24
Support concrete IEEE754 encoding.

makesource:
We finally added concrete IEEE754 encoding to check the performance degradation compared to abstract FP encoding.
This option can be set by giving `--use-concrete-fp` to mlir-tv


aqjune:
Thank you!
I will review this patch today night.

PR #328, opened 2022-08-18
Separate CI image for heaan branch

strikef:
Add a separate image build script for heaan branch.
This will allow non-MLIR API related changes to be applied more easily.

### DIFF
--- a/.github/workflows/env-heaan.yml
+++ b/.github/workflows/env-heaan.yml
@@ -134,9 +134,9 @@ jobs:
       - name: Checkout LLVM
         uses: actions/checkout@v2
         with:
-          repository: llvm/llvm-project
+          repository: aqjune-cryptolab/llvm-project
           path: src/llvm-project
-          ref: 27bb799095c1dfd2e83fe0eabf6399fa83082444
+          ref: heaan-mlir
 
       - name: Get LLVM SHA
         id: sha
--- a/.github/workflows/env-heaan.yml
+++ b/.github/workflows/env-heaan.yml
@@ -1,0 +1,231 @@
+name: Heaan-Image
+
+on:
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  configure-dependencies:
+    if: github.repository == 'aqjune/mlir-tv'
+    runs-on: ubuntu-20.04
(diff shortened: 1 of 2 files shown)


.github/workflows/env-heaan.yml
```diff
@@ -0,0 +1,231 @@
...
+    if: github.repository == 'aqjune/mlir-tv'
+    needs: configure-dependencies
+    runs-on: ubuntu-20.04
+    outputs:
+      sha: ${{ steps.sha.outputs.sha }}
+    steps:
+      - name: Checkout LLVM
+        uses: actions/checkout@v2
+        with:
+          repository: llvm/llvm-project
```
aqjune, line 137: Is this repository path right?

aqjune: We need to clone the custom LLVM.

aqjune:
Build is failing!

.github/workflows/env-heaan.yml
```diff
@@ -0,0 +1,231 @@
...
+          path: src/llvm-project
+          ref: 27bb799095c1dfd2e83fe0eabf6399fa83082444
```
aqjune, line 139: Is it possible to set the LLVM's branch name to `heaan-mlir` instead of designating a commit id?

strikef: Yes, we can use branch name instead of commit hash. Willfix.

.github/workflows/env-heaan.yml
strikef: My mistake :( I'll change the repo path

strikef:
Build is failing because this patch also includes changing the CI image tag that heaan-mlir branch uses. Would it be better to separate it into a separate PR?

.github/workflows/env-heaan.yml
strikef: Fixed

.github/workflows/env-heaan.yml
strikef: Fixed

strikef:
Changing the CI image has been moved to https://github.com/aqjune/mlir-tv/pull/329.
This patch now only adds a workflow to create Heaan-CI image.

PR #329, opened 2022-08-20
Adapt to custom MLIR's API

strikef:
This patch fixes source code, tests, and build script to build against heaan-custom MLIR


strikef:
There are 2 deleted tests (`opts/linalg-canonicalize/from_elem1` and `opts/linalg-canonicalize/from_elem2`).
`linalg.from_elements` at this point did not support creating N-D tensor, so these tests became invalid.

PR #330, opened 2022-08-20
Implement integer extend

strikef:
This PR implements integer extension (`arith.extsi` and `arith.extui`)
* `arith.extsi` is signed integer extend. It transforms an integer from narrower to wider type, by prepending as many 0s(if the MSB of given integer is 0) or 1s(if the MSB is 1) as necessary.
* `arith.extui` is unsigned integer extend. It transforms an integer from narrower to wider type, by prepending as many 0s as necessary.

### DIFF
--- a/src/encode.cpp
+++ b/src/encode.cpp
@@ -733,7 +733,9 @@ void encodeOp(State &st, mlir::arith::ExtSIOp op, bool) {
   }
 
   auto arg = op.getOperand();
-  encodeUnaryOp(st, op, arg, {}, [op_type](Integer &&a) { return a.signedExtend(op_type); });
+  encodeUnaryOp(st, op, arg,
+      {},
+      [op_type](Integer &&a) { return a.signedExtend(op_type); });
 }
 
 template<>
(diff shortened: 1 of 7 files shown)


src/simplevalue.h
```diff
@@ -32,6 +32,9 @@ class Integer {
   static Integer boolTrue();
   static Integer boolFalse();
 
+  Integer signedExtend(const mlir::Type &tgt_type) const;
+  Integer unsignedExtend(const mlir::Type &tgt_type) const;
+
```
aqjune, line 37: Let's not add arithmetic functions to the Integer class; I will do this.

src/encode.cpp
```diff
@@ -716,6 +716,50 @@ void encodeOp(State &st, mlir::arith::ExtFOp op, bool) {
...
+  auto operand_type = op.getOperand().getType();
+  auto src_bw = operand_type.getIntOrFloatBitWidth();
+
+  if (src_bw == tgt_bw) {
+    st.regs.add(op.getResult(), st.regs.get<Integer>(op.getOperand()));
+    return; // extending into identical type is a no-op
+  } else if (src_bw > tgt_bw) {
+    throw UnsupportedException(op.getOperation(),
+      "cannot ExtSI into narrower type!");
+  }
```
aqjune, line 733: extsi and extui syntactically disables these cases; will leave smart_asserts.

PR #332, opened 2022-08-20
Implement bit shift for `smt::Expr`

strikef:
This PR implements `shl`, `ashr`, and `lshr` methods for `smt::Expr`.
* `Expr::shl(const unsigned bits) const` yields a new BV of the same sort with `bits` bits shifted to the left.
* `Expr::ashr(const unsigned bits) const` yields a new BV of the same sort with `bits` bits shifted to the right. Higher bits are filled with 1 if the MSB of the original BV was 1, and 0 otherwise.
* `Expr::lshr(const unsigned bits) const` yields a new BV of the same sort with `bits` bits shifted to the right. Higher bits are always filled with 0.

### DIFF
--- a/src/smt.cpp
+++ b/src/smt.cpp
@@ -1171,7 +1171,7 @@ Expr Expr::shl(const Expr &rhs) const {
   CHECK_LOCK2(rhs);
 
   uint64_t a, b;
-  if (isUInt(a) && rhs.isUInt(b))
+  if (isUInt(a) && rhs.isUInt(b) && b < 64)
     return mkBV(a << b, rhs.bitwidth());
   else if (rhs.isUInt(b)) {
     if (b == 0)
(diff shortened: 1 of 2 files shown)


src/smt.cpp
```diff
@@ -1160,6 +1160,73 @@ Expr &Expr::operator|=(const Expr &rhs) {
...
+  return e;
+}
+
+EXPR_BVOP_UINT64(lshr)
+Expr Expr::lshr(const Expr &rhs) const {
+  CHECK_LOCK2(rhs);
+
+  uint64_t a, b;
+  if (isUInt(a) && rhs.isUInt(b))
+    return mkBV(a >> b, rhs.bitwidth());
```
aqjune, line 1215: Clang lowers `a >> b` into `lshr` and `lshr` produces poison if `b` is equal to or larger than bitwidth (https://llvm.org/docs/LangRef.html#lshr-instruction). I will add `&& b < 64` at the if condition.
Also same for shl.

PR #336, opened 2022-08-29
Implement encoding for scalar function call

strikef:
This PR implements encoding for scalar function call.

As of now, it can only encode very limited set of functions that succeeds to meet the following criterias.
* Function must return scalar value

These restrictions will hopefully be lifted in the further PRs.


src/encode.cpp
```diff
@@ -1,6 +1,9 @@
 #include "encode.h"
 #include "abstractops.h"
+#include "mlir/IR/Types.h"
+#include "mlir/Support/LLVM.h"
```
aqjune, line 4: Let's move these includes to line 7~ below.

tests/litmus/func-ops/scalar.src.mlir
```diff
@@ -0,0 +1,30 @@
+// VERIFY
+
+func.func private @simpl(%v: f32) -> f32
+func.func private @simpl_i(%v: i32) -> i32
+
+func.func @assoc(%v1: f32, %v2: f32) -> f32 {
```
aqjune, line 6: This isn't assoc; it is that add is commutative

src/encode.cpp
```diff
@@ -458,6 +464,140 @@ static void encodeBlock(
...
+    if (getFunction(name)) {
+      // no-op if there already exists a function of the same name
+      return false;
+    } else {
+      calleeMap.insert({string(name),
+                        DeclaredUF::declare(move(domain), move(range), name)});
+      return true;
+    }
+  }
+}
```
aqjune, line 600: Let's split these into a separate header.

src/encode.cpp
```diff
@@ -936,6 +1076,43 @@ void encodeOp(State &st, mlir::arith::SelectOp op, bool) {
...
+      operands.push_back(st.regs.get<Float>(operand));
+    } else if (operandTy.isa<mlir::IntegerType>()) {
+      operands.push_back(st.regs.get<Integer>(operand));
+    } else if (operandTy.isa<mlir::IndexType>()) {
+      operands.push_back(st.regs.get<Index>(operand));
+    }
+  }
+
+  auto calleeUF = *getFunction(callee);
+  st.regs.add(op.getResult(0), calleeUF.apply(operands));
```
aqjune, line 1113: Let's emit a warning message here saying that we assume that call ops are stateless (does not read or write global memory).

src/encode.cpp
strikef: Fixed

tests/litmus/func-ops/scalar.src.mlir
strikef: Fixed. How come I confuse between these two :o

src/encode.cpp
strikef: Moved to `function.h` and `function.cpp`

src/encode.cpp
strikef: Added

strikef: I think we're getting a bit too many warnings (at least 3 times per validation pair). Is this okay, or do you have some better place for the warning?

aqjune: Moved the message to `declareFunction`

src/function.cpp
```diff
@@ -0,0 +1,108 @@
...
+  auto verifComplexity = Complexity::SCALAR;
+  for (const auto &operand_ty : domain) {
+    if (operand_ty.isa<mlir::TensorType>()) {
+      verifComplexity = Complexity::TENSOR;
+    } else if (operand_ty.isa<mlir::MemRefType>()) {
+      verifComplexity = Complexity::MEMREF;
+    } else if (!operand_ty.isIntOrIndexOrFloat()) {
+      throw UnsupportedException("Invalid operand type");
+    }
+  }
```
aqjune, line 32: 1. It is not clear when a function both takes tensor and memref. Or is it assumed that such case cannot happen?

2. What about simply defining two bool variables: `bool hasMemRef` and `bool hasTensor`, and marking a flag to true if any of domain/range is Tensor/MemRef type? I think this will simplify this PR.

strikef: 1. My first insight was that verifying memref(s) will always be more 'complex' than verifying tensor(s), so a function that takes both memref and tensor would be considered as 'memref function'. But now that I think back on it again, it seems that verifying memref and verifying tensor two different tasks, and one's difficulty doesn't necessarily cover the other.
2. Agreed, for the aforementioned reasons. But isn't it only the domain types that affect the difficulty? According to our assumption on UF, we'll get the same value for the same input and different value otherwise, regardless of the range type.

PR #337, opened 2022-09-01
Implement encoding for function calls with tensors

strikef:
This PR lifts some restriction of encoding for scalar function call.

As of now, mlir-tv can encode very limited set of functions that succeeds to meet the following criterias.
* Function must return a scalar value **or a ~~tensor~~ fixed-size tensor** // updated in this PR

More restrictions are to be lifted in the further PRs.


src/function.cpp
```diff
@@ -10,7 +11,16 @@ using namespace smt;
...
 map<string, DeclaredFunction, std::less<>> calleeMap;
+
+vector<uint64_t> getShapeDimVector(const mlir::ShapedType shapedTy) {
+  const auto dims = shapedTy.getShape();
+  return vector<uint64_t>(dims.begin(), dims.end());
+}
+
+mlir::Type getShapeElemType(const mlir::ShapedType shapedTy) {
+  return shapedTy.getElementType();
 }
```
aqjune, line 22: Is this function necessary? this is a one-line function. Let's inline this.

src/function.cpp
```diff
@@ -51,19 +61,27 @@ DeclaredFunction DeclaredFunction::declare(std::vector<mlir::Type> &&domain,
...
+    FnDecl decl(smtDomain, smtRange, string(name) + "_tvfn");
+    return DeclaredFunction(move(domain), move(range), move(decl));
+  } else if (range.isa<mlir::TensorType>()) {
+    const auto tensorRange = range.dyn_cast<mlir::TensorType>();
+    const auto tensorElementType = getShapeElemType(tensorRange);
+    const auto dims = getShapeDimVector(tensorRange);
+    const auto smtRange =
+        static_cast<Expr>(Tensor::var(tensorElementType,
+                                      string(name) + "_tv_ret_tensor", dims))
+            .sort();
```
aqjune, line 75: value.cpp has `arraySortForTensor`.
Let's use it after renaming it to `Tensor::getSort`.

src/function.cpp
```diff
@@ -51,19 +61,27 @@ DeclaredFunction DeclaredFunction::declare(std::vector<mlir::Type> &&domain,
...
-    throw UnsupportedException(
-        "Function that returns tensor is not supported yet");
-  }
-  if (range.isa<mlir::MemRefType>()) {
+  if (range.isIntOrIndexOrFloat()) {
+    const auto smtRange = getScalarSort(range);
+    FnDecl decl(smtDomain, smtRange, string(name) + "_tvfn");
+    return DeclaredFunction(move(domain), move(range), move(decl));
+  } else if (range.isa<mlir::TensorType>()) {
+    const auto tensorRange = range.dyn_cast<mlir::TensorType>();
```
aqjune, line 69: A ranked tensor might be dynamic: `tensor<?x4xi64>`.
Let's raise UnsupportedException in this case.

aqjune: Well, the code update below is checking whether the tensor is unranked...
There are three cases:
1. unranked tensor
2. ranked, dynamically sized tensor
3. ranked, statically sized tensor <-- the only case that must be allowed in this PR
I will update this.

aqjune:
I think this PR can be handily updated to receive ranked static tensors as well. I will update this.

aqjune:
I made some structural changes to DeclaredFunction::declare so that type checks & type conversion to SMT Sorts are done at once.

PR #339, opened 2022-09-02
Fix CI build configurations

strikef:
After recent CMakeFiles refactoring, CI tasks 'build-Z3-only' and 'build-CVC5-only' were building mlir-tv based on wrong configurations.
Added CMake options and fixed CI scripts to fix this issue


aqjune:
Could you describe what was a problem?

strikef:
These builds were supposed to use only one solver each, but all solvers were used in both tasks.

aqjune:
I slightly changed the configuration to explicitly take at least one of USE_Z3 or USE_cvc5.

PR #340, opened 2022-09-05
Re-enable build-release CI

strikef:
The LLVM bug that caused build-release CI to fail [is fixed](https://github.com/llvm/llvm-project/commit/49157fd468521ee482d3f5a0ae1586f9d7c8a550)


aqjune:
Could you update README.md as well? It contains relevant sentences.

PR #341, opened 2022-09-05
Implement encoding for function calls with dynamic tensors

strikef:
This PR lifts some restriction of encoding for scalar function call.

As of now, mlir-tv can encode very limited set of functions that succeeds to meet the following criterias.
* Function must return a scalar value **or a ~~fixed-size~~ ranked tensor** // updated in this PR

More restrictions are to be lifted in the further PRs.

### DIFF
--- a/src/function.cpp
+++ b/src/function.cpp
@@ -51,9 +51,15 @@ DeclaredFunction DeclaredFunction::declare(std::vector<mlir::Type> &&domain,
   };
 
   vector<Sort> smtDomain;
-  smtDomain.reserve(domain.size());
-  transform(domain.cbegin(), domain.cend(), back_inserter(smtDomain),
-            typeToSort);
+  for (const auto operandTy : domain) {
+    smtDomain.push_back(typeToSort(operandTy));
+    if (auto tensorOperandTy = operandTy.dyn_cast<mlir::TensorType>()) {
+      const auto rank = tensorOperandTy.getRank();
+      for (size_t i = 0; i < rank; i++) {
+        smtDomain.push_back(Index::sort());
+      }
+    }
+  }
   FnDecl decl(smtDomain, typeToSort(range), string(name) + "_tvfn");
 
   vector<FnDecl> dims;
@@ -71,7 +77,6 @@ DeclaredFunction DeclaredFunction::declare(std::vector<mlir::Type> &&domain,
 
 ValueTy DeclaredFunction::apply(const std::vector<ValueTy> &operands) const {
   vector<Expr> operandExprs;
-  operandExprs.reserve(operands.size());
 
   auto getZeroGuardedExpr = [](const ValueTy &val) {
     if (holds_alternative<Tensor>(val)) {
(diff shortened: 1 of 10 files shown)


src/function.cpp
```diff
@@ -59,32 +55,82 @@ DeclaredFunction DeclaredFunction::declare(std::vector<mlir::Type> &&domain,
...
+
+      const Expr arr = tensorVal;
+      const auto i = static_cast<Expr>(Index::var("idx", VarType::BOUND));
+
+      Expr zero = Index::zero();
+      const auto elemType = tensorVal.getElemType();
+      if (elemType.isa<mlir::FloatType>()) {
+        zero = Float::constant(llvm::APFloat(0.0), elemType);
+      } else if (elemType.isa<mlir::IntegerType>()) {
+        zero = Integer(0, elemType.getIntOrFloatBitWidth());
```
aqjune, line 92: I remember there was something like `getZero(ty)`... could you check?

src/function.cpp
```diff
@@ -59,32 +55,82 @@ DeclaredFunction DeclaredFunction::declare(std::vector<mlir::Type> &&domain,
...
   vector<Expr> operandExprs;
   operandExprs.reserve(operands.size());
 
+  auto getZeroGuardedExpr = [](const ValueTy &val) {
+    if (holds_alternative<Tensor>(val)) {
+      const auto tensorVal = get<Tensor>(val);
+      Expr numElements = Index::one();
+      for (const auto &dim : tensorVal.getDims()) {
+        numElements = numElements * dim;
+      }
```
aqjune, line 82: I guess there is a function for doing this as well somewhere.

aqjune:
I think the UFs should take the dimension sizes of tensors as arguments too.
For example, this must fail:
```
   %t = 3xi64 zero tensor
   %t2 = tensor.cast %t to ?xi64
   %t3 = call @f(%t2)
   return %t3
=>
   %t = 2xi64 zero tensor
   %t2 = tensor.cast %t to ?xi64
   %t3 = call @f(%t2)
   return %t3
```

src/function.cpp
strikef: We do have `llvm::APFloat::getZero(llvm::Semantics)` but specifying the semantics require several more lines of code. Would it still be better to use getZero?

aqjune: No, there is one is at `value.h`; maybe we can use that.

src/function.cpp
strikef: I first thought of using `std::accumulate`, but `Expr::operator*` is not a binary operator. I also looked for methods like `getNumElements` or sth like that but couldn't find it...

aqjune: I found one - it is `get1DSize` at smt.h. :) You can use it.

strikef: oooooh I see

strikef: Fixed

src/function.cpp
strikef: Fixed

PR #345, opened 2022-09-14
Add command-line option `--use-fn-argument-dims`

strikef:
This PR adds a command-line option `--use-fn-argument-dims`. This option can be used to make an assumption on the shape of the dynamic tensor from a function call.

This option can be used as the following:
* `--use-fn-argument-dims=simpl_fn@0` (the output tensor from `simpl_fn` is assumed to have the same shape as the first argument)
* `--use-fn-argument-dims=fn1@2,fn2@4` (the output tensor from `fn1` is assumed to have the same shape as its third argument, and for `fn2` its fifth argument)

When using this option, UnsupportedException may be thrown under the following circumstances:
* The index used to specify the argument is invalid
* The specified argument or the output type is not shaped
* The output type is unranked or its rank does not match the specified argument
* The shape of the specified argument and the output type does not match
  * Only one of the dimensions on the same rank are dynamic
  * Both dimensions on the same rank are static, but they are not equal

The added tests verify and present its functionalities.

### DIFF
--- a/src/encode.cpp
+++ b/src/encode.cpp
@@ -51,6 +51,27 @@ llvm::cl::opt<bool> arg_use_neg_zero(
   llvm::cl::init(false),
   llvm::cl::cat(MlirTvCategory));
 
+llvm::cl::opt<string> arg_use_arg_dims("use-fn-argument-dims",
+  llvm::cl::desc(
+    "Specify the function argument to use as a reference for the "
+          "output dynamic dims"),
+  llvm::cl::value_desc("<function_name>@idx(,<function_name>@idx)*"),
+  llvm::cl::cat(MlirTvCategory)
+);
+
+namespace {
+  optional<map<string, int64_t, std::less<>>> dimsReferenceIdxMap;
+
+  optional<int64_t> getDimsReferenceIdx(const string_view functionName) {
+    auto itr = dimsReferenceIdxMap->find(functionName);
+    if (itr != dimsReferenceIdxMap->end()) {
+      return itr->second;
+    } else {
+      return nullopt;
+    }
+  }
+}
+
 // map := (i, j, k) -> (j, k, i)
 // input := [a, b, c]
 // output := [b, c, a]
@@ -949,13 +970,44 @@ void encodeOp(State &st, mlir::func::CallOp op, bool) {
       "Invalid number of return values");
   }
 
+  if (!dimsReferenceIdxMap) {
+    // parse user-specified dims
+    dimsReferenceIdxMap = map<string, int64_t, std::less<>>();
+
+    const auto dimRefArgs = arg_use_arg_dims.getValue();
+    if (!dimRefArgs.empty()) {
+      const auto dimRefArgsView = string_view(dimRefArgs);
+      size_t cur = 0;
+      while (true) {
+        // extract name-idx pair
+        const auto commaPos = dimRefArgsView.find(",", cur);
+        const auto dimRefView = dimRefArgsView.substr(cur, commaPos);
+
+        // parse function name and argument index
+        const auto atPos = dimRefView.find("@");
+        auto fnName = string(dimRefView.substr(0, atPos));
+        const auto argIndex = stoll(string(dimRefView.substr(atPos + 1)));
+        auto [_, success] = dimsReferenceIdxMap->insert({move(fnName), argIndex});
+        smart_assert(success, "Dims reference argument for '" << fnName
+                      << "' is specified more than once");
+
+        if (commaPos != dimRefArgsView.npos) {
+          cur = commaPos + 1;
+        } else {
+          break;
+        }
+      }
+    }
+  }
+
   const auto callee = op.getCallee();
   if (!getDeclaredFunction(callee)) {
     vector<mlir::Type> domain(op.getOperandTypes().begin(),
                               op.getOperandTypes().end());
     auto range = op.getResultTypes().front();
     try {
-      declareFunction(move(domain), move(range), move(callee));
+      declareFunction(move(domain), move(range), move(callee),
+                      getDimsReferenceIdx(callee));
     } catch (UnsupportedException e) {
       throw UnsupportedException(op.getOperation(), e.getReason());
     }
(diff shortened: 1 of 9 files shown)


src/encode.cpp
```diff
@@ -51,6 +51,27 @@ llvm::cl::opt<bool> arg_use_neg_zero(
   llvm::cl::init(false),
   llvm::cl::cat(MlirTvCategory));
 
+llvm::cl::opt<string> arg_use_arg_dims("use-fn-argument-dims",
+  llvm::cl::desc(
+    "Specify the function argument to use as a reference for the "
+          "output dynamic dims"),
+  llvm::cl::value_desc("<function_name>@idx(,<function_name>@idx)*"),
+  llvm::cl::cat(MlirTvCategory)
+);
```
aqjune, line 60: You can use `cl::list` with `llvm::cl::MiscFlags::CommaSeparated`.
There will be tools in LLVM or MLIR that use this.

src/function.cpp
```diff
@@ -46,6 +49,31 @@ DeclaredFunction DeclaredFunction::declare(std::vector<mlir::Type> &&domain,
     throw UnsupportedException("Unsupported type: " + to_string(t));
   };
 
+  if (rangeDimRefIdx) {
+    const auto dimRefIdx = *rangeDimRefIdx;
+    smart_assert(dimRefIdx < domain.size(),
+                 "Tried to refer to the argument of an invalid index");
+
+    const auto shapedDimRef = domain[dimRefIdx].dyn_cast<mlir::ShapedType>();
+    const auto shapedRange = range.dyn_cast<mlir::ShapedType>();
```
aqjune, line 58: Is const necessary..?

src/function.cpp
```diff
@@ -46,6 +49,31 @@ DeclaredFunction DeclaredFunction::declare(std::vector<mlir::Type> &&domain,
...
+    smart_assert(
+        shapedDimRef && shapedRange,
+        "Both the specified domain and the range must be shaped types");
+    smart_assert(shapedDimRef.getRank() == shapedRange.getRank(),
+                 "The specified domain and the range must have the same ranks");
```
aqjune, line 63: Let's throw UnsupportedException in this case; assertion failure might look like a bug in mlir-tv.

src/function.cpp
```diff
@@ -141,16 +180,18 @@ optional<DeclaredFunction> getDeclaredFunction(const std::string_view name) {
 }
 
 bool declareFunction(vector<mlir::Type> &&domain, mlir::Type &&range,
-                     const string_view name) {
+                     const string_view name,
```
aqjune, line 183: `const` doesn't seem necessary; let's remove it.

tests/litmus/func-ops/fn_arg_dims.src.mlir
```diff
@@ -0,0 +1,13 @@
...
+func.func private @dyn_tensor_1(%v: tensor<3x?x?xf32>) -> tensor<3x?x?xf32>
+
+func.func @tensor_dim(%v: tensor<3x?x?xf32>) -> i1 {
+  %t = func.call @dyn_tensor_1(%v): (tensor<3x?x?xf32>) -> tensor<3x?x?xf32>
+  %c0 = arith.constant 0: index
+  %d1 = tensor.dim %v, %c0: tensor<3x?x?xf32>
+  %d2 = tensor.dim %t, %c0: tensor<3x?x?xf32>
+  %r = arith.cmpi eq, %d1, %d2 : index
+  return %r: i1
+}
```
aqjune, line 13: Hmm.. it isn't straightforward what these tests do; do you want to update these to simply return a pair of dimension sizes (`tensor.dim`) at 1 and 2 rather than comparisons?

src/function.cpp
strikef: I tend to declare the const variables unless the variable must be mutated. Would it be unnecessary though?

tests/litmus/func-ops/fn_arg_dims.src.mlir
strikef: That makes much more sense. I think I shallow-mindedly copied the test from the other PR :(

src/encode.cpp
strikef: Oooh I wasn't aware of `cl::CommaSeparated` flag. This should make things much easier.

src/function.cpp
aqjune: Hmm, makes sense!

tests/litmus/func-ops/fn_arg_dims_multiple.src.mlir
```diff
@@ -0,0 +1,16 @@
...
+func.func @tensor_dim(%v: tensor<3x?x?xf32>) -> i1 {
+  %f0 = arith.constant 1.0: f32
+  %t1 = func.call @dyn_tensor_1(%f0, %v): (f32, tensor<3x?x?xf32>) -> tensor<3x?x?xf32>
+  %t2 = func.call @dyn_tensor_2(%f0, %f0, %f0, %f0, %v): (f32, f32, f32, f32, tensor<3x?x?xf32>) -> tensor<3x?x?xf32>
+  %c0 = arith.constant 0: index
+  %d1 = tensor.dim %t1, %c0: tensor<3x?x?xf32>
+  %d2 = tensor.dim %t2, %c0: tensor<3x?x?xf32>
+  %r = arith.cmpi eq, %d1, %d2 : index
+  return %r: i1
+}
```
aqjune, line 16: Update this test and below `-bad` one as well?

strikef: Fixed both

tests/litmus/func-ops/fn_arg_dims.src.mlir
strikef: Fixed

src/encode.cpp
strikef: Fixed

src/function.cpp
strikef: Fixed

PR #349, opened 2022-09-20
Encoding for integer shift operations

strikef:
This PR implements encoding for integer shift operations(`arith.shli`, `arith.shrsi`, `arith.shrui`)
* `arith.shli` shifts an integer to the left by a given amount, and fills the 'low-order bits' with 0. `arith.shli %i, %amnt` is equivalent to `arith.mul %i, (math.ipowi 2, %amnt)`
* `arith.shrsi` shifts an integer to the right by a given amount, and fills the 'high-order bits' with the MSB. `arith.shrsi %i, %amnt` is equivalent to `arith.divsi %i, (math.ipowi 2, %amnt)`
* `arith.shrui` shifts an integer to the right by a given amount, and fills the 'high-order bits' with 0. `arith.shrui %i, %amnt` is equivalent to `arith.divui %i, (math.ipowi 2, %amnt)`

If the shift amount is greater than or equal to the integer's bitwidth, it *seems to be* UB. We don't have any official documentation to back up this claim though. I'll undraft this PR once I have a stronger evidence.

### DIFF
--- a/src/encode.cpp
+++ b/src/encode.cpp
@@ -773,7 +773,8 @@ void encodeOp(State &st, mlir::arith::ExtUIOp op, bool) {
       [extamnt](Integer &&a) { return ((Expr)a).zext(extamnt); });
 }
 
-template <> void encodeOp(State &st, mlir::arith::ShLIOp op, bool) {
+template <>
+void encodeOp(State &st, mlir::arith::ShLIOp op, bool) {
   auto arg = op.getOperand(0);
   const auto src_bw = arg.getType().getIntOrFloatBitWidth();
   auto shlamnt = op.getOperand(1);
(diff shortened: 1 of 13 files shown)


src/encode.cpp
```diff
@@ -71,6 +71,43 @@ namespace {
       return nullopt;
     }
   }
+
+  void encodeShiftAmountBound(State &st, mlir::Operation* op) {
+    const auto arg = op->getOperand(0);
+    const auto argTy = arg.getType();
+    const auto amnt = op->getOperand(1);
+    const auto amntTy = amnt.getType();
```
aqjune, line 79: I think variable `arg` is not necessary because (1) UB depends on the shift amount only, and (2) amntTy == argTy. What do you think?

aqjune: Then let's leave `smart_assert` to check whether types are equal.

PR #352, opened 2022-10-22
Upgrade mlir

Hatsunespica:
I tried to upgrade the project to the newest MLIR 0w0.
Now it passes complication. 
However `Linalg` dialect removed `InitTensorOp` thus all related tests should be updated.  I tried to update them but I am not quite sure about finding a substitute op. Maybe [tensor.empty](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorempty-mlirtensoremptyop) ? 

Thanks!

### DIFF
--- a/src/encode.cpp
+++ b/src/encode.cpp
@@ -1589,6 +1589,8 @@ void encodeOp(State &st, mlir::tosa::AvgPool2dOp op, bool) {
           "Zero-padded pooling is supported only.");
   }
 
+  // TODO: The current modeling ignores the acc_type attribute.
+
   auto result = input.avgPool(kernelDims, strides);
   st.regs.add(op.getResult(), move(result));
   st.wellDefined(op, input.isFullyInitialized(), "source tensor initialized");
--- a/src/encode.cpp
+++ b/src/encode.cpp
@@ -897,7 +897,7 @@ void encodeOp(State &st, mlir::arith::TruncIOp op, bool) {
 
 template<>
 void encodeOp(State &st, mlir::linalg::IndexOp op, bool) {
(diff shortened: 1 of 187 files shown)


aqjune:
Hi @Hatsunespica !

Yes, tensor.empty seems okay. I am on a travel right now; I will look into this when I am back.

Hatsunespica:
Hi Seunghyeon,
Thanks for your investigation!
I'll do it

All the best,
Yuyou

Seunghyeon Nam ***@***.***> 于2022年10月24日周一 08:03写道：

aqjune:
Oops, there are a few other failures - do you want to have a look?

src/encode.cpp
```diff
@@ -3662,7 +3636,6 @@ static void encodeBlock(
     ENCODE(st, op, mlir::linalg::FillOp, encodeMemWriteOps);
     ENCODE(st, op, mlir::linalg::GenericOp, encodeMemWriteOps);
     ENCODE(st, op, mlir::linalg::IndexOp, encodeMemWriteOps);
-    ENCODE(st, op, mlir::linalg::InitTensorOp, encodeMemWriteOps);
```
aqjune, line 3665: It is slightly sad that this `ENCODE` is permanently being removed.
Do you think it is feasible to add
```
ENCODE(st, op, mlir::tensor::EmptyOp, encodeMemWriteOps);
```
again? It might require a very small change to the existing `encodeOp` for `linalg::InitTensorOp`.

aqjune:
Thank you for providing the code. :) I updated the code & test and now it passes CI successfully.

